PR #4, opened 2015-07-14
react-devtools -> redux-devtools

kevinrenskers:


gaearon:
lol thanks

erikras:
I also lol'd at this, but dammit if I haven't mistyped it about seven times since then. It's a shame that both libraries start with the same two characters.

PR #7, opened 2015-07-14
Clarify how to exclude DevTools in production

gaearon:


erikras:
Or how to ***NOT*** use DevTools in production. :smile: 

:beer: Cheers for the readme link.

Issue #3, opened 2015-07-14
Challenge: Implement a cool custom monitor


gaearon:
`<DevTools>` accepts a `monitor` prop. The monitor that ships with `react-devtools` is called `LogMonitor` but it's just a React component. You can build your own instead.

I challenge you to implement a really different monitor than what we ship with. Some crazy ideas:

* A slider that lets you jump between computed states just by dragging it
* An in-app layer that shows the last N states right in the app (e.g. for animation)
* A time machine like interface where the last N states of your app reside on different Z layers

[`LogMonitor` propTypes](https://github.com/gaearon/redux-devtools/blob/a831af7317340af66b2fe638bf6e974dd83fc480/src/react/LogMonitor.js#L5-L13) should give you some ideas about the stuff injected by DevTools:

```js
static propTypes = {
  // Stuff you can use
  computedStates: PropTypes.array.isRequired,
  currentStateIndex: PropTypes.number.isRequired,
  stagedActions: PropTypes.array.isRequired,
  skippedActions: PropTypes.object.isRequired,

  // Stuff you can do
  reset: PropTypes.func.isRequired,
  commit: PropTypes.func.isRequired,
  rollback: PropTypes.func.isRequired,
  sweep: PropTypes.func.isRequired,
  toggleAction: PropTypes.func.isRequired, // ({ index })
  jumpToState: PropTypes.func.isRequired // ({ index })
}
```

This data (and functions) will be injected into your monitor by the dev tools. Feel free to use (or ignore) any of them, and let me know if something important you need for a custom monitor is missing.

gaearon:
This example commit is a good starting point for adding DevTools to your React project: https://github.com/gaearon/redux-devtools/commit/0a2a97556e252bfad822ca438923774bc8b932a4. (Nevermind the Webpack config changes.)

To work on a custom monitor, you can remove `DebugPanel` and `LogMonitor` completely and instead use `<DevTools store={store} monitor={MyCustomMonitor}>`.

gaearon:
To make it easier to build a slider-like monitor, I added two new props to the monitor contract: `jumpToState({ index })` and `currentStateIndex`. Use them to implement slider time travel interface without extra hassle.

pburtchaell:
One thing I would find really helpful in the default LogMonitor is a shortcut to show/hide the dev tools. Something like `ctrl+opt+h` maybe. If you like the idea, I can make a PR.

hzoo:
[...]

PR #12, opened 2015-07-19
Implement toggle

pburtchaell:
- Adds `monitorState` object to track the state of the monitor, e.g., if it is visible or not.
- Adds `show()` action creator to show monitor
- Adds `hide()` action creator to hide monitor
- Moves the padding from the monitor container to the monitor itself; having the padding on the container prevented the monitor from being fully hidden.
- Adds some help text at the top of the monitor to explain how to hide the monitor.

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -86,7 +86,7 @@ function liftReducer(reducer, initialState) {
     skippedActions: {},
     currentStateIndex: 0,
     monitorState: {
-      isVisible: true,
+      isVisible: true
     }
   };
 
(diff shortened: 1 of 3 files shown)


gaearon:
Thank you!!

gaearon:
I think we can probably generalize `show` and `hide` as something like `setUIState` and `this.props.uiState` so different monitors can use `devTools.js` but store different things there.

pburtchaell:
Yeah, that would be a good idea. Perhaps something to mention on #3.

PR #13, opened 2015-07-19
Update examples with README and correct URLs

pburtchaell:
- Adds README file to each example
- Uses the redux-devtools URL in package.json


gaearon:
:+1: I think running `npm i` at the root is also required, isn't it?

Issue #14, opened 2015-07-19
'Ctrl + h' not working on Chrome + Linux


rodrigobac:
I searched a little and just discovered that Chrome always return false to event.ctrlKey and other keys (e.g. event.shiftKey) on keypress.

So, the hide function doesn't work on Chrome + Linux (34+ as far as I know).

pburtchaell:
Interesting. Did anyone mention a work-around? I have implemented key combinations like this before and haven't had any issues in Chrome.

rodrigobac:
I'll look for workarounds (Having "ctrl" shotcuts in my actual project would be great). But, so far, none of them worked.

rodrigobac:
I've managed to fix by just changing the 'keypress' event to 'keydown' event (yeah, hours looking for workarounds and just tried that now).

pburtchaell:
Cool. Looks like it was merged in so you should be :100: here.

gaearon:
Out in 0.1.2, thank you a lot.

PR #17, opened 2015-07-20
Fixes Issue #14

rodrigobac:
- Changed keypress event to keydownd event. In Chrome + Linux, the keypress event don't return ctrlKey event. ( Fixes issue #14 );
- Added event.preventDefault() to prevent default behavior from Browsers;
- Changed keyCode 8 to 72 to properly refer to 'h' key


pburtchaell:
:+1:

PR #25, opened 2015-07-23
Changed cursor: hand to pointer

taylorhakes:
`cursor: hand` doesn't work in Chrome extensions and is not the correct value for current browsers. I updated `hand` to use `pointer`


gaearon:
Thank you!

gaearon:
Should be released in 0.1.3.

PR #26, opened 2015-07-23
Fixed bug in sweep

taylorhakes:
Sweep was throwing an error because of an index out of range issue.

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -129,7 +129,7 @@ function liftReducer(reducer, initialState) {
     case ActionTypes.SWEEP:
       stagedActions = stagedActions.filter((_, i) => !skippedActions[i]);
       skippedActions = {};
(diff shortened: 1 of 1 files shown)


gaearon:
Interesting why it happened. By that moment the actions are already filtered so shouldn't it point to the last state?

taylorhakes:
I am a little confused at what you are asking. The `currentStateIndex` before this action occurs is normally pointing to the last element (before filter). The elements are then filtered, removing the ones that are skipped which makes the array shorter (smaller length). If you do `Math.max` it keeps the `currentStateIndex` the same, outside the bounds of the array. `Math.min` is needed to make sure the `currentStateIndex` is inside the bounds of the new array.

gaearon:
Lol I'm bad at math. Thank you for explaining.

gaearon:
Should be released in 0.1.3.

PR #27, opened 2015-07-23
Re: Challenge: Implement a cool custom monitor (#3)

theaqua:
I saw #3 and try bring UX for monitor.

* less unused space
* JSON formatted for better readability

Before:
image

After:
image


hmeerlo:
Nice, especially the readability

gaearon:
Thanks for the effort! In the end we went with #48 because it has a tree inspector.

theaqua:
@gaearon yep, awesome job.

PR #29, opened 2015-07-24
initial test for the counter and todomvc examples

knowbody:


gaearon:
I think you meant to do this in Redux repo? Anyway, we have tests there now so I'm closing.

knowbody:
yep, forgot to close that one @gaearon

PR #30, opened 2015-07-27
reducer exception shown in console

FredyC:
Error thrown in reducer is consumed and only message is shown. This PR adds call to `console.error` to see a stack trace.


gaearon:
:+1: sorry I left that out!

gaearon:
Should be released in 0.1.3.

PR #35, opened 2015-07-31
Add timestamp to recorded actions

calesce:
addressing #3, where timestamps were requested for recorded (or lifted) actions, for the purpose of making a "video player" of actions possible.

I made a "factory" of sorts for the ```INIT``` action, so resets/commits/etc will all have a timestamp from start to finish. That might be overkill, since a "replayer" could just start with the action after ```INIT```.

Also, I know I'm directly mutating the lifted actions when adding a timestamp property, so let me know if there's a more "pure" way to do that. I also (slightly) namespaced the timestamp property, in case user actions have a "timestamp" property already, which I can see leading to some nasty bugs.

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -76,7 +76,6 @@ function recomputeStates(reducer, committedState, stagedActions, skippedActions)
   return computedStates;
 }
 
-
 /**
  * Lifts the app state reducer into a DevTools state reducer.
  */
@@ -88,7 +87,8 @@ function liftReducer(reducer, initialState) {
     currentStateIndex: 0,
     monitorState: {
       isVisible: true
-    }
+    },
+    timestamps: [Date.now()]
   };
 
   /**
@@ -101,7 +101,8 @@ function liftReducer(reducer, initialState) {
       skippedActions,
       computedStates,
       currentStateIndex,
-      monitorState
+      monitorState,
+      timestamps
     } = liftedState;
 
     switch (liftedAction.type) {
@@ -110,17 +111,20 @@ function liftReducer(reducer, initialState) {
       stagedActions = [INIT_ACTION];
       skippedActions = {};
       currentStateIndex = 0;
+      timestamps = [liftedAction.timestamp];
       break;
     case ActionTypes.COMMIT:
       committedState = computedStates[currentStateIndex].state;
       stagedActions = [INIT_ACTION];
       skippedActions = {};
       currentStateIndex = 0;
+      timestamps = [liftedAction.timestamp];
       break;
     case ActionTypes.ROLLBACK:
       stagedActions = [INIT_ACTION];
       skippedActions = {};
       currentStateIndex = 0;
+      timestamps = [liftedAction.timestamp];
       break;
     case ActionTypes.TOGGLE_ACTION:
       skippedActions = toggle(skippedActions, liftedAction.index);
@@ -130,6 +134,7 @@ function liftReducer(reducer, initialState) {
       break;
     case ActionTypes.SWEEP:
       stagedActions = stagedActions.filter((_, i) => !skippedActions[i]);
+      timestamps = timestamps.filter((_, i) => !skippedActions[i]);
       skippedActions = {};
       currentStateIndex = Math.min(currentStateIndex, stagedActions.length - 1);
       break;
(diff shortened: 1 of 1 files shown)


src/devTools.js
```diff
@@ -136,6 +139,7 @@ function liftReducer(reducer, initialState) {
       if (currentStateIndex === stagedActions.length - 1) {
         currentStateIndex++;
       }
+      liftedAction.action.devTimestamp = Date.now();
```
gaearon: Shouldn't this be part of `liftedAction` itself? This is what it's for: to carry the information that makes sense to the lifted reducer. 

How about changing

```js
function liftAction(action) {
  const liftedAction = { type: ActionTypes.PERFORM_ACTION, action };
  return liftedAction;
```

to

```js
function liftAction(action) {
  const liftedAction = {
    type: ActionTypes.PERFORM_ACTION,
    timestamp: Date.now(),
    action
  };
  return liftedAction;
```

instead?

If you do it this way, you can change the lifted reducer to extract timestamps into a separate `actionTimestamps` lifted state field, which dev tools can access.

calesce: Yeah, that's cleaner. The way I currently have it implemented still attaches the timestamp, but doing so in the reducer makes less sense.

In the monitor I'm working on, I basically ignore the idea of toggling actions, so my thought was that I would just use the stagedActions array and grab the timestamp for the action matching the current state index (as well as the next action if not the last). Haven't had time to try that out yet, so maybe it wouldn't work.

gaearon: OK, let me know how it goes!

gaearon:
Is this already used somewhere? I'm confused because `INIT` action is not a lifted one. Also, you have not provided access to timestamps from the state, as they are not handled by the reducer. How would a monitor read it?

calesce:
My thought was that timestamps could be accessed by ```stagedActions```, which is why I wanted to add a timestamp to created ```INIT``` actions as well. That would make a replay in real time possible for each action that's been dispatched. I guess that could alternatively be done for ```computedStates```.

calesce:
OK, looking at your previous comment (8 days ago) I think adding an ```actionTimestamps``` field for the reducer to pass to the monitor would be a better approach, rather than having the monitor look through timestamps in ```stagedActions```.

src/devTools.js
```diff
@@ -138,6 +143,7 @@ function liftReducer(reducer, initialState) {
         currentStateIndex++;
       }
       stagedActions = [...stagedActions, liftedAction.action];
+      timestamps = [...timestamps, Date.now()];
       break;
```
gaearon: I still think it's better to put timestamp into liftedAction and grab it from there. This way reducer is completely deterministic.

calesce: Gotcha. For RESET, COMMIT, ROLLBACK, where the timestamp array is reset, do you think it's better to put the timestamp in their action creators as well? Since ```liftAction()``` is a de facto action creator so that would make it a consistent pattern.

gaearon: They wouldn't need timestamps. We only want timestamp alongside user's actions. Therefore they should be assigned in `liftAction()` and 

```js
       stagedActions = [...stagedActions, liftedAction.action];
      timestamps = [...timestamps, Date.now()];
```

would be 

```js
       stagedActions = [...stagedActions, liftedAction.action];
       timestamps = [...timestamps, liftedAction.timestamp];
```

Even if you put timestamps on COMMIT, etc, there is no code that would read them.

calesce: I wanted there to be a new starting point for timestamps after RESET, etc, similar to the INIT action.

gaearon: Aaargh, I lost track of what you mean :-) Can you try explaining again?

My suggestion was to record the time in `liftAction` because that corresponds precisely to when it fired in the app.

calesce: No worries, this is just prevalent for me because I want to work with it in the slider monitor. :smile:

Yeah, I agree recording timestamps in ```liftAction``` is good. I just want to be able to start a replay from app startup or RESET, etc, which would mean recording timestamps in those instances. That way, the ```timestamps``` array is always is corollated with ```computedStates``` and ```stagedActions```. Hope that makes sense.

gaearon: I just understood what you mean. Yes. Go for it :-)

calesce: Ok, done!

gaearon:
OK, this looks good to me. Have you tried it with the slider? We don't have tests yet so manual testing is needed.

calesce:
no I haven't. I've just verified that timestamps looks right at different points in time via ```console.log``` debugging :wink: on both LogMonitor and the slider.

I haven't figured out the implementation for real-time replay yet, but I'll have time to tonight (7-8 hours from now).

calesce:
yay, thanks! I'll follow up on the replay stuff.

Issue #34, opened 2015-07-29
devTools should have a way to remove stagedActions


ioss:
One UseCase: After having called jumpToState a user might want to remove all actions following that state.

At the moment this would be pretty complicated for a Monitor. Either modifying stagedActions directly and calling jumpToState again to have recomputeStates run. 
Or store the currently skipped Actions, reset the skipped actions and only skip the actions after the current action and then call sweep and then restore the remembered skipped ones... ;)

Also - not sure -, maybe there should be a recomputeStates action.

gaearon:
I agree! Something like `spliceActions` would be nice.

gaearon:
Seems like we already support this with `RECOMPUTE_STATES` (to be changed to `IMPORT_STATE` soon).

PR #39, opened 2015-08-05
"splicing" actions

calesce:
This is a somewhat quick-and-dirty approach to #34. I first went with the suggested "spliceActions" approach, but I realized I'd have to keep track of the lifted action and then manually dispatch ```performAction``` with somewhere else.

Also, this doesn't do much to address ```skippedActions```.

I imagine that in most cases, if you want to fire new actions after calling ```jumpToState```, you would want the action to update the state currently being viewed, rather than jumping to the end of the log and appending there. However, I can understand if this doesn't seem like a universal-enough use case to where you'd merge it.

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -135,8 +135,12 @@ function liftReducer(reducer, initialState) {
     case ActionTypes.PERFORM_ACTION:
       if (currentStateIndex === stagedActions.length - 1) {
         currentStateIndex++;
+        stagedActions = [...stagedActions, liftedAction.action];
+      } else {
+        stagedActions = stagedActions.slice(0, currentStateIndex + 1).concat(liftedAction.action);
+        currentStateIndex = stagedActions.length - 1;
+        skippedActions = {};
       }
-      stagedActions = [...stagedActions, liftedAction.action];
       break;
     case ActionTypes.SET_MONITOR_STATE:
       monitorState = liftedAction.monitorState;


ioss:
I think we need more general/basic actions, so that a monitor can decide what to do with the staged actions.
Or we allow for monitors to call recomputeState (still don't like that).

I am in favor of more general but specified actions as well as in specified action-history structures (you could have branches of staged actions you can switch between for example).

Maybe we should create another issue to discuss the additional Actions and Action-DevStates?

calesce:
I actually kind of like the idea of allowing monitors to use a ```recomputeStates``` action with its choice of stagedActions, committedState, etc. That way you can paste in actions/states from separate debug sessions, which makes something like [this](https://www.youtube.com/watch?v=5yHFTN-_mOo) possible.

I imagine the action would look something like:
```javascript
{
  type: 'RECOMPUTE_STATES',
  stagedActions,
  skippedActions,
  committedState,
}
```

Thoughts?

ioss:
While this is nice to "hack in" anything you might ever want. It does not help with interoperability of Monitors.

I think Actions like 
ADDED, REMOVED, HIDDEN, 
UPDATED (to add monitor specific properties to the liftet action),
CLONED_STAGED (to save multiple action history pathes), 
SWITCHED_TO_STAGED/CLONE, 
and so on

with a tighter contract will make working with different monitors at the same time possible, without every monitor having to know the requirements and functionality of every other monitor there is. ;)

Also we should question skippedActions, this should imho be a property (skipped/filtered) of the lifted action, and maybe have some other lifted action states (visible/hidden, ...)

Maybe just rename this Issue to "DevTools specification discussion"?

calesce:
Yeah you should open a new issue for that (or re-use #34), not sure this PR is the right thread. I do agree that there should be more discussion on this kind of stuff. I also agree on skippedActions, you can't really splice in skippedActions easily if you want to change stagedActions.

I'm not sure that doing RECOMPUTE_STATES is as damaging as you suggest. Of course it's up to the monitor to use it properly. But if the user can only manipulate one monitor at a time, then as long as that monitor produces proper actions/states then that's okay, the other monitors will have the proper UI (because it's React! :smile:).

There's a lot more thought to be had here, it's all very exciting to me! Multiple branches of actions is an interesting thought as well.

calesce:
After #49 this can now be implemented in monitors, feel free to close if you don't want this to be default behavior.

gaearon:
@calesce @ioss 

I got confused now and forgot everything we discussed. What should I do with this PR and https://github.com/gaearon/redux-devtools/pull/62? Are they solving different needs?

ioss:
@gaearon I had to check myself....
(This PR changes the way actions are appended, after you jumped to a previous action.)

While this PR/CR is different from #62, I think the new RECOMPUTE_STATES (or should it be called IMPORT_STATES?) - containing your fix 62e5316, followed by my fix #62 for the actionCreator - makes this PR irrelevant, as the new RECOMPUTE_STATES makes it possible for every monitor to do what ever fits its UX best. 

You could say that with the more powerful RECOMPUTE_STATE most of the other actions can be emulated - so I would leave them untouched, required different behavior should be monitor-userland now, until we have a better picture how monitors use recomputeState.

Hope this helps?

calesce:
Yeah, I believe this can be replicated using the ```recomputeStates``` action creator. The purpose here was to start a new timeline of actions from ```currentStateIndex``` if the ```currentStateIndex``` was somewhere in the "past" and a new action was lifted. I'd still be interested in seeing it merged, but it doesn't need to be.

ioss:
@calesce: After thinking a little bit more about it, I agree, this should be merged. (As it actually can not be easily emulated by `recomputeState`, as the deletion of the following actions happens at PERFORM_ACTION, so you are able to jump around as long as no further action is issued.)

As the new action is created in a state at the currentStateIndex, the current procedure to appending it to the end (or inserting it - even worse) is invalid.

calesce:
Cool. If @gaearon agrees I'll update the PR.

gaearon:
Actually I just realized what's happening here, and I don't think this is right approach. It's one of the original design decisions I made with these devtools: they must be consistent. The fact that you are *looking* at past state does not excuse you to fork the history. If you'd like to fork, please do it explicitly (and thus erase the future). Otherwise, you'd hit small inconsistencies all the time. For example, an async action success or a periodic action may come after you have travelled back in time, and it no longer applies cleanly.

gaearon:
Let's track this here: #118

Issue #45, opened 2015-08-10
Idea: Monitor for React Perf


gaearon:
React has perf method to measure wasted renders. Idea: a monitor that calls `React.addons.Perf.start()` and `stop()` before and after replaying staged actions, and shows the results in table. Every time you hot reload components, it would do this again, so you see how your changes are affecting performance.

I'm not sure it's easy to implement as is, because React Hot Loader won't tell you if hot reload has happened when the component handled it, but it's something we want to consider in the future.

Thansk @winkler1 for the tip.

winkler1:
Woohooh! Some good "ambient awareness" ideas from the stack overflow guys -- Miniprofiler puts an subtle page render time on the screen. Geeks probably can't help glancing at it and trying to optimize it. https://youtu.be/fN37OvBNZOk?t=43

The CSS looks good and is designed for reuse - https://github.com/MiniProfiler/ui

gaearon:
Closing for inactivity.

PR #48, opened 2015-08-10
Improved design and tree for devtools

dzannotti:
* Improved the overall design 
* Rollback/sweep/commit now stands on the top of devtools
* You can now pass any base16 theme name prop or your own base16 colors to `<LogMonitor/>` (defaulting to ocean)
* Payload and state are expandable tree with variable type coloring
* States automatically highlights the differences between current and previous state with a subtle red background

Here is a screenshot of what it looks like using ocean's theme
Original PR:
screen shot 2015-08-10 at 18 56 26

With changes:
screen shot 2015-08-10 at 23 54 11


gaearon:
Some screenshots would be nice ;-)

FredyC:
@gaearon Just wanted to say the same thing, you were faster. Might just add that there is another try to make it look better at #36. I suppose that these two are not compatible :)

gaearon:
I think it's a bit more ambitious than #36

dzannotti:
screen shot 2015-08-10 at 18 56 26
Here's it what it currently looks like

FredyC:
Well I am simply amazed. Definitely step forward compared to the current simple thing :) Although doesn't make much sense to me to display `3 Items` for an simple object. It looks somewhat confusing to me.

Also it might be nice to have an option to just collapse whole action (perhaps even by default) instead of just disabling it.

Another neat idea that just struck me is about display time of an action since we have it in there now.

dzannotti:
Updated description of features in original post

dzannotti:
@FredyC 3 items is displayed only when the tree is not expanded (useful to know how many array items or object keys you have in state)

gaearon:
I want some people who use DevTools to give it a go on a real project before merging and let us know any point points. Any volunteers?

FredyC:
@dzannotti Yeah I am not sure if knowing number of keys is useful anywhere :) Perhaps you can change that label to display "3 keys" instead? This way it's easily confused with array and adds more complexity your brains has to pay attention to.

dzannotti:
Collapsing the whole action is something i did think about today, but the ui to do that would require additional things (ie fontawesome?) which would complicate the distribution of the package, unsure about this one.
Regading the time, it doesn't makes much sense since you can replay your action, perhaps would be useful to time each reducer instead - probably related to https://github.com/gaearon/redux-devtools/issues/45

FredyC:
@gaearon Yeah I would love to give it a try, but I am now in a part of the project where Redux isn't used, so I don't volunteer :)

dzannotti:
+1 what @gaearon says, i use devtools on my current project, and built this around my thoughts/needs, would be interested what the others says about it

gaearon:
Would it be better to introduce some vertical margin between actions? Right now it's not immediately obvious that trees "belong" to the header above.

VinSpee:
Looking nice! I think it would also be valuable to make the "Reset" "Revert" "Sweep" and "Commit" actions look more like buttons, as right now it's not clear that you can act on them.

dzannotti:
@gaearon @vinspee maybe we can rotate the base16 palette up a notch and solve both, will experiment

damassi:
I would increase the inner vertical padding within each action and knock small text size down 1-2 points.  And if you're going for a coder font, use Monaco.

oal:
I think it looks very good, but there is too little contrast, especially between the red ADD_TODO and the background.

dzannotti:
@oal the color picking it's done thru base16, you can define a new set and pass it as props and i will be used

oal:
@dzannotti How about changing the default to something with more contrast? Like base16-bright?

dzannotti:
@oal: I'm assuming you are referring to action type coloring, that one is picked based on a action type (so each action has their own coloring) between base7 and baseF, the background is base0/base1.

@damassi which vertical padding are you referring to?

davidbehrend:
From a design perspective I'd love to see the nested arrows line up with their parent title.

Like this:
```
----------------------------
@@INIT
----------------------------
▼ state: {} 1 Item
  ▶ todos: [] 1 Item
----------------------------
```

As opposed to this:
```
----------------------------
@@INIT
----------------------------
▼ state: {} 1 Item
 ▶ todos: [] 1 Item
----------------------------
```

cleercode:
If you're going to use a monospace font, ensuring that the characters are actually lined up would be an easy design win. Currently that's being thrown off by slightly different font sizes for the keys and values. Letter spacing on lowercase monospace text also seems unnecessary.

jjt:
Looking great!

@dzannotti Wouldn't need Fontawesome for collapsing the actions. You could just use the same arrows that you do for the other trees.

Agreed on the contrast. Looks like the action type bg is base2, which is too bright for a lot of the base-16 themes, Ocean included. Maybe switch to base0 for both payload and state, and use base1 for the action type bg? IMO, don't need zebra striping for just 2 items :)

damassi:
@dzannotti - I'm referring to the rows holding the `ACTION_NAME`s -- if they were a bit taller, while maintaining the same text size and keeping everything vertically centered, I think it would look a bit nicer.

webpro:
Looking very useful overall :-)

* Sometimes a state tree collapses as a whole if I click some sub key.
* The "reset" etc. buttons seem inactive. Is this expected, or should I enable something?
* These labels on top where not very clear to be a button.
* A draggable width (or height if horizontal?) for the bar would be super-useful. Or at least configurable width (or can I already do this?)

nicinabox:
**Contrast** 

I think this has already been touched on, but the contrast seems a bit low, especially the red-on-gray. It's not immediately clear to me what the colors indicate, but would a white or light gray work instead?

**Type**

The condensed titles and the tracked out toggles are distracting. The letter spacing is probably unnecessary. They typefaces similar, but different enough to feel "off" from each other. Key/value pairs should use the same font size. A single typeface should be sufficient.

**Indents**

Like @davidbehrend mentioned, nested arrow alignment would be nice.

**Buttons**

I had no trouble understanding that Reset/Revert/Sweep/Commit were actions. It would be easy to overdesign these.

Overall, pretty good. Just needs a little balance.

dzannotti:
@webpro: re the reset button, to do something you need more than one state non commited, try to add a todo then push reset

gaearon:
I get `Unknown node type: undefined` occasionally in the console.

gaearon:
Maybe change `payload` to `action`? That's how we call it everywhere in the docs.

dzannotti:
@gaearon: can you give me a state/payload to reproduce that? That error was a safety net for when  Object.prototype.toString.call(obj).slice(8, -1) returns undefined

dzannotti:
Updated changed screenshot, I have tried to bump op the theming by 1 but it looked wrong, in the end i found the middle ground (by removing devtools opacity), @oal @jjt @nicinabox

jjt:
Removing the opacity helped, but still a bit hard to read the red `ADD_TODO`, IMO. Heavier font weight would probably help.

nicinabox:
@jjt is that Monaco?

dzannotti:
@nicinabox yes it is
@jjt i tried bolding it out, but didn't woek very well

jjt:
Thanks for trying that out. I guess I'll just use a different theme :)

nicinabox:
Sometimes these things are hard to put into words, so I made you a mockup.

* Swapped Monaco for Roboto
* Swapped Courier (?) for Monaco
* More whitespace
* More obvious indentation and alignment
* Higher contrast (base16 Bright, but same background colors from @dzannotti's mockup)
* Small things like pluralization and string quoting (a la chrome dev tools)
* Left out title colors because I didn't understand it

devtools

gaearon:
*Oh.* That's beautiful.

dzannotti:
@nicinabox woaaaa thanks! the arrows don't line up (on purpose?) anymore
Can you list out which base16 colors you used please so i can put them in straight?
About the colors for action type, was a @gaearon request to recognize actions at glance.
About the fonting, i love it, but would mean adding an external reference to google fonts (not sure how @gaearon feels about this, but i would like devtools to work offline too.
Great work, looks beautiful

jjt:
That looks better to my eyes, esp the white titles.

Could make a little coloured box, or gravatar-type thing to differentiate between the action types.

gaearon:
>About the colors for action type, was a @gaearon request to recognize actions at glance.

I think it's now legible enough that I don't care about colors.

>About the fonting, i love it, but would mean adding an external reference to google fonts 

Hm.. Any way we can serve them locally? I'm feeling really dumb right now..

nicinabox:
> the arrows don't line up (on purpose?) anymore

Eh, it's the unicode triangle. Could probably find something better. 

> Can you list out which base16 colors you used please so i can put them in straight?

Post these in a sec.

> About the fonting, i love it, but would mean adding an external reference to google fonts
I assumed they'd be bundled? Monaco is only available on OSX I believe. Presumably if you're serving up styles, fonts should be possible too?

dzannotti:
@gaearon: best possible option would be to configure webpack to support fonts from `./node_modules/redux-devtools` and then a css from there too

gaearon:
@dzannotti Does this mean user has to do something, or would it work out of the box?

dzannotti:
@gaearon: assuming we can serve the css with css-modules inside dist, to the best of my knowledge, it means the user has to serve a ttf file from node_modules

nicinabox:
```
Background            #2A2F3A
Title background      #4F5A65
Title                 #FFFFFF
Muted title           #5A6673
Key                   #6FB3D2
Object value          #BEBEBE
Array value           #D381C3
String value          #A1C659
Number value          #6FB3D2
Boolean value         #6FB3D2
```

dzannotti:
@nicinabox any chance of the base16 values for it? 
something like
```
Background base00
Title background base02
``
etc...
the ones i'm using on the default theme are  from https://github.com/chriskempson/base16-builder/blob/master/schemes/ocean.yml
all of the themes are viewable here http://chriskempson.github.io/base16/

**Comment by nicinabox** [comment]
Ah, I didn't use all the ones from Base16 Bright unfortunately. The ones I'm missing are from your original (except for the muted title which is lower brightness). Do they need to all be from the same set?

Background               #2A2F3A
Title background         #4F5A65
Title              07    #FFFFFF
Muted title              #5A6673
Key                0D    #6FB3D2
Object value       03    #BEBEBE
Array value        0E    #D381C3
String value       0B    #A1C659
Number value       09    #6FB3D2
Boolean value      09    #6FB3D2

**Comment by dzannotti** [comment]
@nicinabox possibly yes, would enable all base16 themes

**Comment by nicinabox** [comment]
Here we go

```
Background         00
Title background   02
Title              06
Muted title        02
Muted background   01
Key                0D
Object value       03
Array value        0E
String value       0B
Number value       09
Boolean value      09
```

And full on Bright colors:

devtools

VinSpee:
I still really don't think there's a strong enough affordance on the actions. Not clearly interactable. 



(null)

dzannotti:
awesome, ty @nicinabox!

nicinabox:
@VinSpee do you have a suggestion? I've tried it here as an "action bar" at the bottom instead. If they don't look actionable, what do they look like? If you were trying do something actionable, where would you look?

devtools

dzannotti:
There's a hover where the buttons go darker by the way which isn't clear from the screenshots

nicinabox:
I think the top actions are better. You'll be more likely to notice them at a glance as a top-down reader. Bottom actions might be very far away considering monitor and window sizes. Not only would you have to move the mouse far, but I suspect the likelihood of noticing them goes way down.

mjw56:
Is it possible to make the click area a little more fine grained for toggling the action? It seems when toggling the arrow next to the action I keep accidentally cancelling that action.

image

dzannotti:
@mjw56 you should be able to click on the key/value to expand/collapse

mjw56:
@dzannotti yeah, it is working and looking great. Only, there seems to be a small space between the arrow and action word where when you click there, it will cause to cancel that action?

VinSpee:
I was thinking something along these lines for the buttons @nicinabox:

proposed

top or bottom, I don't have a strong opinion.

nicinabox:
Maybe a color is enough? I don't particularly like the visual clutter of the buttons and divisions if whitespace is enough to suggest what they are.

devtools
devtools - buttons
devtools - dividers

VinSpee:
I think the third option (with separators) may suffice. Without some sort of clear outline, it's not obvious that they're interactable.

nicinabox:
Yeah, I'm not sure. Apple and Google have been pushing borderless buttons with success. My question is still: if they don't look actionable, what do they look like? IMO they're too random to be mistaken for titles. 

It would only take an accidental mouseover or quick exploration to see they're actionable if you were confused about what they were there for. The labels also read as an action, which I think give a lot of clarity to what they do rather than applying extra visual style.

Maybe it's worth a/b testing if that's a point of confusion for folks?

nicinabox:
FWIW just asked my girlfriend which of the three looked the most clickable. Her responses:

1. Maybe?
2. Definitely.
3. Maybe?

VinSpee:
Simple is better :-)



(null)

nicinabox:
Here's tweaked buttons showing hover state on the first one.

devtools - buttons

VinSpee:
I dig it! Awesome work on this. 



(null)

FredyC:
I agree that buttons looks good like this, but what about disabled state for those? Like reset if there is nothing to reset actually. Is it enough to just take out hover? Shouldn't it be more obvious what you can use and what not? I am not UX, I am just talking out of my feeling.

FredyC:
Also I have noticed another issue with showing number of keys together with ImmutableJS structures.

    const M = Immutable.Map({ A: 1, B: 2, C: 3 });
    Object.prototype.toString(M); // -> [object Object]
    Object.keys(M); // -> ["size", "_root", "__ownerID", "__hash", "__altered"]

I am not sure how this should be solved actually. It would require some better detection mechanism than just `toString`.

dzannotti:
@fredyc parse(stringify(state)) should fix this, will do this and the theme/design changes when I get back home later

FredyC:
@dzannotti Unfortunately that doesn't work for a List. 

    const original = Immutable.List([1, 2, 3]);
    const converted = JSON.parse(JSON.stringify(original));
    Object.prototype.toString(converted); // -> [object Object]

FredyC:
@dzannotti There will a slight hickup for Immutable.List even with that approach.

    const original = Immutable.List([1, 2, 3]);
    const converted = JSON.parse(JSON.stringify(original));
    Object.prototype.toString(converted); // -> [object Object]

So array would be considered an object and rendered using `JSONObjectNode` thus displaying keys and values to be equal.

gaearon:
>agree that buttons looks good like this, but what about disabled state for those? 

Can't we keep them like now? Hide them when they're disabled?

gaearon:
I have a gut feeling stringifying and parsing Immutable will be pretty slow. I'd rather have official support for it via duck typing.

davidbehrend:
I'd suggest avoiding 'pure black' (#000).

Ian Storm Taylor has a good [blog post](http://http://ianstormtaylor.com/design-tip-never-use-black/) on the reasoning why.

FredyC:
> Can't we keep them like now? Hide them when they're disabled?

Hiding buttons is bad approach in my opinion and there is perhaps some UX rule for that. For example you have intention to click once, but your finger gets crazy and clicks twice suddenly clicking on something else.

gaearon:
Good point, thanks.

nicinabox:
> I'd suggest avoiding 'pure black' (#000).

I totally agree (and my [original](https://github.com/gaearon/redux-devtools/pull/48#issuecomment-129652708) used a blue shade instead). However, if @dzannotti wants to support the base16 themes we lose the flexibility on that. The specific color can fluctuate a lot between themes. 

> I agree that buttons looks good like this, but what about disabled state for those?

I wasn't sure if a disabled state existed, but now watching @gaearon's gif I see that it does. Should be easy enough. Something like this?

screen shot 2015-08-11 at 7 38 09 am

@dzannotti that's `02` for the disabled color and no background.

VinSpee:
Looks great!

iamdustan:
There’s a lot of comments so I may have missed this while glancing through, but has there been any consideration of publishing this LogMonitor standalone? Leave the default one a bit lacking to encourage explorations such as this one in the community and make it easier for folks to try this out today.

damassi:
@iamdustan - *very* much agree with your comment

dzannotti:
i don't know how i feel about @iamdustan's comment on github, i sort of agree, but i sort of want to see the official devtools to be the best they could be (and i actually wouldn't want to limit the PR to this improvement either, if more improvements are accepted)

nicinabox:
Needs a bit more work, but this was a good first pass I think.

devtools mov

dzannotti:
+1 on @nicinabox theme used as default and his tweaking

gaearon:
Anything else here? I can't wait to merge. :-)

gaearon:
Awesome job. I'm so grateful to you @dzannotti @nicinabox.

iamdustan:
:clap::clap: as are we all. This is awesome.

FredyC:
Looks like I am too late in here. Seems you forgot about that support for Immutable. I didn't actually tried, I just don't see where it's taken care of. Might be just looking at wrong places, in that case nevermind :) And good job, looks good (at least from the picture)

gaearon:
Out in 1.0

gaearon:
Please file an issue for supporting Immutable.

nicinabox:
Thanks guys! I added #52 to continue polishing. If you spot issues just give a shout.

PR #49, opened 2015-08-10
Update devtools.js to make it able recompute states on action

lapanoid:
We think that this change is essential to make [redux-import-export-monitor](https://github.com/lapanoid/redux-import-export-monitor)

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -6,7 +6,8 @@ const ActionTypes = {
   SWEEP: 'SWEEP',
   TOGGLE_ACTION: 'TOGGLE_ACTION',
   JUMP_TO_STATE: 'JUMP_TO_STATE',
-  SET_MONITOR_STATE: 'SET_MONITOR_STATE'
+  SET_MONITOR_STATE: 'SET_MONITOR_STATE',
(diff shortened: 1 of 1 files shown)


dzannotti:
That is cool, perhaps it should be part of the official tools too?

lapanoid:
If @gaearon decide so it would be cool or we can keep it in separate repo either way is fine.

lapanoid:
I just think import/export should be a minimal monitor so it could be composed with other monitors. It should not tightly coupled with LogMonitor.

gaearon:
Looks good, thanks.

ioss:
With RECOMPUTE_STATES there might be a problem (well, actually this was a problem before too, but not as tempting) of monitors mutating the state by forgetting to clone the state - this probably would not be a problem for the mutating monitor, but for others. 
Should this be prevented by devTools or do we trust the monitors?

gaearon:
We'll trust the monitors for now. It's all experimentation so we're relying on monitor devs knowing what they're doing.

ioss:
Ok. I think this can be closed then. (Eventually after the minor fix for: 62e5316)

Issue #52, opened 2015-08-12
Finish styling


nicinabox:
A checklist for tracking/finishing styling.

* [x] box-shadow still visible after hiding devtools
* [x] in hidden state, DebugPanel is still visible (https://github.com/gaearon/redux-devtools/issues/54)
* [ ] add hover state on LogMonitorEntryAction title
* [ ] add vertical margin on deeply nested values
* [x] change error color to red (#59)
* [x] does action section need more callout?
* [ ] animate appending items (@threepointone?)
* [ ] key may be difficult to expand (I noticed the actions block has an odd conflict with the mouse)

gaearon:
Currently action contents is highlighted because of a little bug. It's the same highlight that highlights the state changes. However I decided to release that as is, because it's actually better (to me) when the action is highlighted. After all it's the *reason* something changes. We can fix the bug, but then I think action section should have a lighter background so there's more accent on it.

What do you think?

gaearon:
I was also wondering if you'd like to consider using [react-motion](https://github.com/chenglou/react-motion) for some animations. For example a very subtle "add" animation could help distinguish just inserted actions. I wonder if @threepointone is interested in experimenting with something like this.

nicinabox:
My thoughts were that action was already priority by nature of being first. I changed the whitespace around the content sections so the next section (`state`) is farther away, giving action a tad more weight. With my short exploration last night the background seemed to add more noise than clarity. 

Reminds me of that quote:

> If you make everything bold, nothing is bold

Right now all the backgrounds have at least one reason for existing:

* Title backgrounds function as a visual divider and denote (more subtly) an action
* Button backgrounds are a visual divider and action
* Highlight backgrounds show a change in a sea of content that's difficult to cross reference without it

An action content background says "look at me!", but provides no significant improvement in usability, imo.

gaearon:
Fair. Would you like me to fix the bug then?

nicinabox:
+1 animations would be a nice touch. It's a little hard to tell you're at the bottom when a new item its appended. I keep scrolling the wrong direction.

gaearon:
Idea: we can add “previous value” tooltip to changed primitive values on hover.

nicinabox:
I like it.

nicinabox:
Or what about like git style? Cross out the previous value, show in red. Show the new value next to it in green. Then you can do away with the extra background too.

gaearon:
[...]

PR #57, opened 2015-08-12
lightweight css animations for new actions entering the log.

threepointone:
while we ponder react-motion in #52, consider a simpler alternative for now :) this PR adds a css fadeIn animation to entries in the actions log.

### DIFF
--- a/examples/counter/index.html
+++ b/examples/counter/index.html
@@ -1,6 +1,18 @@
 <html>
   <head>
     <title>Redux Counter Example</title>
+    <style>
+      /* unsure how to represent this in js, so it stays here for now */
+      @-webkit-keyframes fadeIn {
+        from { opacity: 0; }
+        100% { opacity: 1; }
+      }
+
+      @keyframes fadeIn {
+        from { opacity: 0; }
+        100% { opacity: 1; }
+      }
+    </style>
   </head>
   <body>
     <div id="root">
--- a/src/react/LogMonitorEntry.js
+++ b/src/react/LogMonitorEntry.js
@@ -9,6 +9,16 @@ const styles = {
   },
   tree: {
     paddingLeft: 0
+  },
+  animated: {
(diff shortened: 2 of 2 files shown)


nicinabox:
Would be awesome if you could show a gif of it!

threepointone:
it's quite subtle, doesn't show well (and seems longer) in this gif http://recordit.co/ss6t7FyU7H :|

nicinabox:
Nice! What happens when you get down to the bottom and it starts scrolling?

threepointone:
just adds it at the bottom as it already does now, along with the subtle fade. try it, it's really not much :)

gaearon:
I'll defer any visual decisions to @nicinabox to approve

nicinabox:
How do we feel about the flash? For reference:

fade mov

gaearon:
What if we the opposite? Start off actions brighter than usual, then let them “cool down”. Like changing nodes in Chrome inspector.

nicinabox:
That's a neat effect too. I think the biggest thing that's lacking with each of these is you don't really see the content move upwards (but these are still an improvement).

flash mov

dzannotti:
i like all 3

nicinabox:
Oops, probably wasn't clear on the first one. It's @threepointone's. I just recorded at 30fps and showed scrolling.

nicinabox:
Could we append it *then* animate scrolling to the bottom?

gaearon:
Yeah. Let's replace naïve scrolling with something like https://github.com/alexkuz/additive-animation#usage-example?

gaearon:
Closing, as I don't currently have time to make it better, and scrolling seems to be important for the desired effect. If someone would like to implement it, please do it in https://github.com/gaearon/redux-devtools-log-monitor instead (track https://github.com/gaearon/redux-devtools/pull/132 for discussion).

PR #62, opened 2015-08-13
add parameters to ActionCreator.recomputeStates - esp. timestamps.

ioss:
timestamps were used in the reducers, but were missing on the ActionCreator.
Added timestamp and optional currentStateIndex & skippedActions
Added tests: you might want to move them into their own files?

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -152,8 +152,11 @@ function liftReducer(reducer, initialState) {
       stagedActions = liftedAction.stagedActions;
       timestamps = liftedAction.timestamps;
       committedState = liftedAction.committedState;
-      currentStateIndex = stagedActions.length - 1;
-      skippedActions = {};
+      currentStateIndex = liftedAction.currentStateIndex;
+      if (currentStateIndex === -1) {
+        currentStateIndex = stagedActions.length - 1;
+      }
+      skippedActions = liftedAction.skippedActions;
       break;
     default:
       break;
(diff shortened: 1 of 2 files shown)


src/devTools.js
```diff
@@ -247,11 +250,14 @@ export const ActionCreators = {
   setMonitorState(monitorState) {
     return { type: ActionTypes.SET_MONITOR_STATE, monitorState };
   },
-  recomputeStates(committedState, stagedActions) {
+  recomputeStates(committedState, stagedActions, timestamps, currentStateIndex=-1, skippedActions={}) {
```
gaearon: Can you describe the use case for this in more detail? How do you imagine this being used?

ioss: Main use case was for import after export (check why it was introduced in the first place). At the moment if you export and then import you will loose timestamps - if you use the ActionCreator.recomputeStates and do not create your Action by "hand".

Also I think it might be nice to have a monitor which stores multiple "actionsequences" and provide a way to switch between them without using the clipboard - which is just export/import again using for example localstorage/sessionstorage...

Additionally you will loose the currentIndex and the skipped actions, this is not that important, as you can restore these by dispatching multiple toggle + one jumpto action, but if we open up the state anyway, why not provide these optionally.

ioss: Probably this should be refactored to recomputeStates(devState) or better import(devState), getting rid of all these parameters and in that way making sure, that future additions to the devState will create an api change.

ioss: See refactoring at https://github.com/gaearon/redux-devtools/compare/master...ioss:fix-for-recompute-states-2

gaearon:
I'm going to close this and instead implement something close to #118.
Unless I'm wrong, these are mostly equivalent, but `IMPORT`/`EXPORT` explains what happens better.

ioss:
They are completely equivalent, and I like IMPORT/EXPORT better (that's why I did the second one)

PR #65, opened 2015-08-13
added flexDirection on menu items

bloodyowl:
as people may use a facebook/css-layout like reset that makes
flexDirection default to "column".

screen shot 2015-08-13 at 13 05 32


nicinabox:
:+1:

Issue #51, opened 2015-08-12
Support for Immutable.js


adamalbrecht:
Since upgrading to 1.0, support for Immutable.js has stopped working. It shows the raw immutable object rather than converting it to JS.

There is some discussion of this in #48

![screen shot 2015-08-12 at 10 13 57 am](https://cloud.githubusercontent.com/assets/180203/9226072/e81aac8a-40da-11e5-95b0-a59c3e5ca6d5.png)

FredyC:
Ah, so you were faster than me to report this. Thanks :) As @gaearon mentioned, this should be solved with duck typing, perhaps checking if there is a `toJS` method and invoking it before passing for any checks.

hugobessaa:
I hit the same problem, but I'm using `mori` for my data-structures. I think the DevTools component could receive `converter`, a function that would be applied as a map function for the whole state.

```
<DebugPanel top right bottom>
  <DevTools 
    store={store}
    converter={mori.toJs}
    monitor={LogMonitor} />
</DebugPanel>

**Comment by FredyC** [comment]
@hugobessaa I see that even mori implements iterator interface, so there is really no need for a converter. We just need someone willing to implement (see #66).

**Comment by hugobessaa** [comment]
Oh, I see. Nice.

**Comment by weltenwort** [comment]
For everyone who's waiting for that pull request to be merged, here's a small workaround via the `select` property of the `DevTools` component:
```jsx
let selectDevToolsState = (state = {}) => Immutable.fromJS(state).toJS();

<DebugPanel top right bottom key="debugPanel">
    <DevTools store={store} select={selectDevToolsState} monitor={LogMonitor} />
</DebugPanel>
```

It's a bit crude, but does the job for now.

gaearon:
Should be fixed in 1.1.0, please verify.

weltenwort:
:+1: works for me, thank you!

PR #66, opened 2015-08-13
Add support for Map nodes

bnoon:
I use Map() in my state for ordered lists and would like to inspect them.

### DIFF
--- a/src/react/JSONTree/JSONMapNode.js
+++ b/src/react/JSONTree/JSONMapNode.js
@@ -1,0 +1,124 @@
+import React from 'react';
+import reactMixin from 'react-mixin';
+import { ExpandedStateHandlerMixin } from './mixins';
+import JSONArrow from './JSONArrow';
+import grabNode from './grab-node';
+
+const styles = {
+  base: {
+    position: 'relative',
+    paddingTop: 3,
+    paddingBottom: 3,
+    marginLeft: 14
+  },
+  label: {
+    margin: 0,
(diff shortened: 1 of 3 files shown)


FredyC:
I think that this should be sorted out in a more generic way, since there is a lack of support for Immutable (see #51). The Map has method `toJSON` same as the ImmutableJS objects. If we use duck typing for such method, we don't need additional node type in my opinion.

bnoon:
I agree that the copy/paste was a quick hack and the common functionality should be pulled out.

However, Map is a built-in object and sufficiently different from Object or Array to be special cased.

FredyC:
Yeah, I guess that Map is different, but it's pretty much same as Map from Immutable. There is problem that you can have basically anything in your key, so it cannot be displayed easily in a tree view. Map can be more likely seen as a list of entries with keys and values.

gaearon:
Can we somehow use iterators + `entries()` method to have a generic way to enumerate stuff?

FredyC:
@gaearon Yeah, I think that might be a good way. The result is in both cases identical (https://jsbin.com/pirura/edit?js,console). And actually `entries()` is default iterator. It might be enough to check if there is an iterator present `obj[Symbol.iterator]`.

bnoon:
I am closing this in favor of #79.  Thanks @FredyC!

gaearon:
Should be fixed in 1.1.0, please verify.

PR #67, opened 2015-08-14
DefinePlugin works

forresto:
https://webpack.github.io/docs/list-of-plugins.html#defineplugin

'ProvidePlugin' didn't work.


gaearon:
Ouch, my mistake. Thanks. :+1:

PR #70, opened 2015-08-15
LogMonitor now attempts to call .toJS() on state entries.

davezuko:
This is an attempt to fix the unreadable state entries for ImmutableJS. I really didn't want to do the manipulation in the render method, but from what I saw in terms of how states are computed and stored I couldn't think of a better solution that didn't significantly alter that part of the codebase.

Anyways, feedback is appreciated. Maybe somebody has a better solution in mind, but this seems to be working for me.

### DIFF
--- a/src/react/LogMonitor.js
+++ b/src/react/LogMonitor.js
@@ -144,6 +144,20 @@ export default class LogMonitor {
       const action = stagedActions[i];
       const { state, error } = computedStates[i];
       let previousState;
+
+      const readableState = {};
+      for (const prop in state) {
+        if (
+          state.hasOwnProperty(prop) &&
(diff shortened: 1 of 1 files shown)


davezuko:
Aaaaand just saw https://github.com/gaearon/redux-devtools/pull/66 so this might not be the desired solution. I'll read through that again and see if I can rework the implementation.

gaearon:
It's best if we can support anything iterable, not just Immutable.

davezuko:
Right, totally makes sense. I'll try to update the PR later when I get home.

gaearon:
Closing in favor of #79.

gaearon:
Should be fixed in 1.1.0, please verify.

PR #75, opened 2015-08-18
add visibleOnLoad props to LogMonitor

jackielii:
I'm not sure whether to use `componentWillMount` or `componentDidMount`. But both seems to work, and `componentWillMount` won't trigger another `render`.

But are we changing props indirectly here right? Is this the right way to go?

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,9 @@ A live-editing time travel environment for [Redux](https://github.com/rackt/redu
 * If you change the reducer code, each “staged” action will be re-evaluted
 * If the reducers throw, you will see during which action this happened, and what the error was
 * With `persistState()` store enhancer, you can persist debug sessions across page reloads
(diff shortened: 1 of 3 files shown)


rockingskier:
Hi @jackielii looks like a useful feature. Just thought I'd point out that you've spelt it `visiableOnLoad` rather than `visibleOnLoad`

jackielii:
@rockingskier thanks! I thought it was a bit wired when I spelled it out...

nadavspi:
:+1:

Graf009:
+1

gaearon:
I think we should keep it open in the examples.
Instead, we can document `visibleOnLoad` in README.

gaearon:
There is a conflict. Can I ask you to rebase on master, and then squash all commits into a single commit?

jackielii:
done.

tedpennings:
:+1:

dtryon:
:+1:

gaearon:
Should be out in 2.1.0.

PR #78, opened 2015-08-19
custom parsers for persistState

FredyC:
Since I am using ImmutableJS for my state, the current `persistState` cannot work properly there, eg. my state is List of Record instances. This cannot be solved in some generic way. So I made this modification which allows me to pass this object as a second argument to `persistState` middleware.

    const TGDocumentParser = {
        serialize(state) {
            return state.toJS();
        },
        deserialize(state) {
            return Imm.List(state.map((row) => new TGRow(row)));
        }
    };

    persistState(sessionId, {
        tgDocument: TGDocumentParser
    });

Each state slice can have a different parser. The actual serialization isn't implemented yet and I am wondering it it's needed. Immutable can serialize just fine. I am not sure about other structures like mori or es6. What do you think?

### DIFF
--- a/src/persistState.js
+++ b/src/persistState.js
@@ -1,6 +1,50 @@
-export default function persistState(sessionId) {
+export default function persistState(sessionId, parsers) {
   if (!sessionId) {
     return next => (...args) => next(...args);
+  }
+
+  function callParser(parserMethod, parserName, stateSlice) {
+    const parser = parsers[parserName];
+    if (parser && typeof parser[parserMethod] === 'function') {
+      return parser[parserMethod](stateSlice);
+    }
+    return stateSlice;
+  }
+
+  // function serializeState(state) {
+  //   if (!state) {
+  //     return state;
+  //   }
+  //   return Object.keys(state).reduce((serialized, sliceName) => {
+  //     serialized[sliceName] = callParser('serialize', sliceName, state[sliceName]);
+  //     return serialized;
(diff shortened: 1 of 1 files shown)


FredyC:
@gaearon Any thoughts here?

FredyC:
@gaearon, @quirinpa @bnoon : Any thoughts about this?

gaearon:
I'd aim for less API, is this possible? Can we just give it `reviver` argument a la `JSON.parse(json, reviver)`?

FredyC:
@gaearon I am not sure that's possible. The reviver is called for every (nested) key in there. This cannot be handled in generic way, because you need to know your state structure to be able transform it to correct objects.

gaearon:
With the API you propose, is it only possible to choose the reviver for the topmost state slices? Is it assumed that everything below is homogenous (lol at this word)?

FredyC:
Alright, so instead of passing those parsers in there, it can be just a function that gets called for every whole state. That might work better I guess. I will try.

FredyC:
Alright, I think it's really better. There is a less code and it can handle various cases much better. My deserializer function just changed a bit.

    function deserializer(state) {
        return {
            ...state,
            tgDocument: Imm.List(state.tgDocument.map((row) => new TGRow(row)))
        };
    }

    persistState(sessionId, deserializer);

gaearon:
Should be out in 2.1.0.

PR #79, opened 2015-08-20
support for iterable structures within state

### LINKED ISSUE #51: Support for Immutable.js

FredyC:
Alright, I finally needed this so badly that I had to do it :) I haven't tested all available structures, but here is view of my Immutable.List of Immutable.Record instances within state.

redux-devtools-iterable

Fixes #51, #66, #70

### DIFF
--- a/src/react/JSONTree/JSONIterableNode.js
+++ b/src/react/JSONTree/JSONIterableNode.js
@@ -1,0 +1,141 @@
+import React from 'react';
+import reactMixin from 'react-mixin';
+import { ExpandedStateHandlerMixin } from './mixins';
+import JSONArrow from './JSONArrow';
+import grabNode from './grab-node';
+
+const styles = {
+  base: {
+    position: 'relative',
+    paddingTop: 3,
+    paddingBottom: 3,
+    paddingRight: 0,
+    marginLeft: 14
+  },
+  label: {
+    margin: 0,
+    padding: 0,
+    display: 'inline-block'
+  },
+  span: {
+    cursor: 'default'
+  },
+  spanType: {
+    marginLeft: 5,
+    marginRight: 5
+  }
+};
+
+@reactMixin.decorate(ExpandedStateHandlerMixin)
+export default class JSONIterableNode extends React.Component {
+  defaultProps = {
+    data: [],
+    initialExpanded: false
+  };
+
+  // flag to see if we still need to render our child nodes
+  needsChildNodes = true;
+
+  // cache store for our child nodes
+  renderedChildren = [];
+
+  // cache store for the number of items string we display
+  itemString = false;
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      expanded: this.props.initialExpanded,
+      createdChildNodes: false
+    };
+  }
+
+  // Returns the child nodes for each element in the array. If we have
+  // generated them previously, we return from cache, otherwise we create
+  // them.
+  getChildNodes() {
+    if (this.state.expanded && this.needsChildNodes) {
+      let childNodes = [];
+      for (const entry of this.props.data) {
+        let key = null;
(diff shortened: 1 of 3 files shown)


gaearon:
Good stuff, thank you! Anybody else wants to test drive this?

quirinpa:
I think i do eventually at least

bnoon:
With a couple of changes, this will work for `Map()` as well, so I am closing #66 in favor of this more general solution.

src/react/JSONTree/JSONIterableNode.js
```diff
@@ -0,0 +1,141 @@
...
+  getItemString() {
+    if (!this.itemString) {
+      const { data } = this.props;
+      let count = 0;
+      if (typeof data.count === 'function') {
+        count = data.count();
+      } else {
+        count = data.length;
+      }
+      this.itemString = count + ' entr' + (count !== 1 ? 'ies' : 'y');
```
bnoon: Change to the following
``` javascript
      if (typeof data.count === 'function') {
        count = data.count();
      } else if (typeof data.size !== 'undefined') {
        count = data.size;
      } else {
        count = data.length;
      }
```
To handle the length of a `Map()`

src/react/JSONTree/obj-type.js
```diff
@@ -1,3 +1,7 @@
 export default function(obj) {
-  return Object.prototype.toString.call(obj).slice(8, -1);
+  const type = Object.prototype.toString.call(obj).slice(8, -1);
+  if (type === 'Object' && typeof obj[Symbol.iterator] === 'function') {
```
bnoon: When using the babel `Map` polyfill, the `.toString()` returns 'Map'.

Change to
```
  if ((type === 'Object' || type === 'Map') && typeof obj[Symbol.iterator] === 'function') {
```

FredyC: I am not sure this is correct approach. There might be even more iterable types returning its own toString representation. 

Originally I had there just check for iterator existence, but that was wrong for String which implements iterator for its characters. Similar goes for the Array.

It might be safer bet to just check `typeof obj === 'object'` and excluding array from it. Any other ideas?

src/react/JSONTree/JSONIterableNode.js
FredyC: Yeah, I agree. I had the `size` there before, but then I've noticed that MoriJS has only `count` method.

I am thinking if it wouldn't be better to use iterator even here to count entries. That would be unified way. I would like to avoid adding another property/method in a future if some other data structure implementing iterator would use different name for this.

src/react/JSONTree/obj-type.js
gaearon: >It might be safer bet to just check typeof obj === 'object' and excluding array from it. Any other ideas?

Fair.

src/react/JSONTree/JSONIterableNode.js
gaearon: Any conclusion here?

FredyC: Alright, I've change it. It uses the `size` property as it's the most common (only mori doesn't have it), otherwise it counts entries through the iterator. I am not using `length` because it's deprecated in Immutable anyway and others doesn't have it at all.

FredyC: Another small change to use Number.isSafeInteger for the `size` property. It should be _safer_ way in case some library has `size` as a function.

gaearon:
Out in 1.1.0, let me know how it goes!

PR #82, opened 2015-08-20
Improve instructions, adding examples and notes on middleware ordering

qnm:
After discussion in gaearon/redux-thunk#9 I've prepared slightly more elaborate installation instructions based on my current understanding of installation.

Please let me know if I've got anything wrong!

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ A live-editing time travel environment for Redux.
 npm install --save-dev redux-devtools
 ```
 
-DevTools is a middleware, which should be added to your middleware stack *after* `applyMiddleware` because `applyMiddleware` is potentially asynchronous. Otherwise, DevTools won't see the raw actions emitted by the Promise middleware etc.
+DevTools is a store enhancer, which should be added to your middleware stack *after* `applyMiddleware` as `applyMiddleware` is potentially asynchronous. Otherwise, DevTools won't see the raw actions emitted by the Promise store enhancer etc.
 
 To install, firstly import `devTools` into your App container:
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,6 +21,47 @@ A live-editing time travel environment for Redux.
 npm install --save-dev redux-devtools
 ```
 
+DevTools is a middleware, which should be added to your middleware stack *after* `applyMiddleware` because `applyMiddleware` is potentially asynchronous. Otherwise, DevTools won't see the raw actions emitted by the Promise middleware etc.
```
gaearon: >DevTools is a middleware

It's a “store enhancer”. The result of `applyMiddleware` is also a “store enhancer”.

README.md
```diff
@@ -21,6 +21,47 @@ A live-editing time travel environment for Redux.
...
+Lastly, include the `DebugPanel` in your page:
+```
+export default class App extends Component {
+  render() {
+    return (
+      <Provider store={store}>
+        {() => <CounterApp />}
+      </Provider>
+      <div>
+        <Provider store={store}>
```
gaearon: This looks like extra copy-paste, we don't need `<CounterApp>` twice.

qnm:
@gaearon Sure. Updated.

gaearon:
Sorry, I confused you all over. :-P

“Middleware” is the thing that wraps `dispatch`. [redux-thunk](https://github.com/gaearon/redux-thunk), [redux-promise](https://github.com/acdlite/redux-promise) are all middleware.

“Store enhancer” is the thing that wraps `createStore`. Redux DevTools and `applyMiddleware` are both store enhancers.

Hope it makes more sense.

qnm:
@gaearon Doesn't this need further changes, or will you make them yourself?  I was just about to pick this up again!

gaearon:
I just tweaked it a bit. Thanks!

Issue #83, opened 2015-08-23
JSONObjectNode#getChildNodes: Cannot read property '...' of null (previousData)


sompylasar:
First reported by @bkniffler [here](https://github.com/gaearon/redux-devtools/issues/52#issuecomment-132236190) and by @josebalius [here](https://github.com/gaearon/redux-devtools/issues/52#issuecomment-133118028). Just hit this issue, too. 

The error happens on [this line](https://github.com/gaearon/redux-devtools/blob/master/src/react/JSONTree/JSONObjectNode.js#L62) when clicking a state object node in the devtools to expand it.

gaearon:
We're happy to accept a PR, looks like `previousData` may not be there. If you do a PR, please make similar adjustments for other nodes in the codebase (you can grep “previousData”).

josebalius:
@gaearon I submitted https://github.com/gaearon/redux-devtools/pull/84 to fix this. I added a null check everywhere previousData was checking for undefined.

gaearon:
Should be fixed in 1.1.1.

PR #84, opened 2015-08-23
Fix for issue #83 - JSONObjectNode#getChildNodes: Cannot read propert…

josebalius:
…y '...' of null (previousData)


gaearon:
Out in 1.1.1, thank you!

Issue #85, opened 2015-08-24
Split JSONTree into its own npm package?


chibicode:
Hi, I'm wondering if `JSONTree` can be its own npm package. I need a JSON viewer for my app, and I thought it'd be nice if I can use this.
https://github.com/gaearon/redux-devtools/blob/master/src/react/JSONTree/index.js

Of course I can just import `redux-devtools/src/react/JSONTree` but it'd be nicer if it's in own npm package. I can make a PR if that's ok.

For now, I think I'm going to use this:
https://github.com/Lapple/react-json-inspector

gaearon:
Please make a PR. We want to split `LogMonitor` itself too: https://github.com/gaearon/redux-devtools/issues/61. Splitting JSONTree is a nice start.

Please make sure to keep [the credits](https://github.com/gaearon/redux-devtools/blob/master/src/react/JSONTree/index.js#L1-L4) somewhere in `package.json` and/or README:

```js
// ES6 + inline style port of JSONViewer https://bitbucket.org/davevedder/react-json-viewer/
// all credits and original code to the author
// Dave Vedder <veddermatic@gmail.com> http://www.eskimospy.com/
// port by Daniele Zannotti http://www.github.com/dzannotti <dzannotti@me.com>
```

chibicode:
@gaearon thanks! Made a pull request: https://github.com/gaearon/redux-devtools/pull/86

gaearon:
Awesome, out in 1.1.2.

PR #86, opened 2015-08-24
Extract JSONTree

chibicode:
- Closes https://github.com/gaearon/redux-devtools/issues/85
- JSONTree has been extracted to [chibicode/react-json-tree](https://github.com/chibicode/react-json-tree) ([npm](https://www.npmjs.com/package/react-json-tree)).
- Tested by running `examples/counter` and `examples/todomvc` against it.
- `src/utils/deepEqual.js` and `src/utils/hexToRgb.js` were both only used by JSONTree, so I removed those as well.
- [The original credits](https://github.com/gaearon/redux-devtools/blob/master/src/react/JSONTree/index.js#L1-L4) are moved to [README.md](https://github.com/chibicode/react-json-tree/blob/master/README.md#credits) and [package.json](https://github.com/chibicode/react-json-tree/blob/74734c0e9dd369ddfed90bc22f134af2629bb412/package.json#L23-L26).
- Don't mind transferring the ownership of [chibicode/react-json-tree](https://github.com/chibicode/react-json-tree) to @gaearon. Just let me know.


chibicode:
**Note:** I realized that `JSONTree` actually doesn't work if `theme` prop is missing.

As an emergency fix, I copied over `solarized.js` and made it as the default theme for `JSONTree` if its `theme` prop is missing. This way, users don't have to set the themes themselves for most use cases.

I chose Solarized (even though it's not the default for [redux-devtools](https://github.com/gaearon/redux-devtools) because I think most people will use JSONTree standalone on a white background (it's also the default browser background color), and I found that Solarized is most legible for white backgrounds.

I think the longer term solution would be to maybe pull out [redux-devtools's themes](https://github.com/gaearon/redux-devtools/tree/master/src/react/themes) directory into its own package and require it on both `LogMonitor` and [react-json-tree](https://github.com/chibicode/react-json-tree). What do you think?

[Here's the commit](https://github.com/chibicode/react-json-tree/commit/cc0c187ed10ca35442872d11e5e4558e04e8335e) for adding a default theme. The screenshot is below (this is also on the [README](https://github.com/chibicode/react-json-tree/blob/master/README.md) of [react-json-tree](https://github.com/chibicode/react-json-tree).)

---



---

Finally, I also [added an example](https://github.com/chibicode/react-json-tree/commit/941f750685cb5f182105353916aa4142b19ee0fd) for `JSONTree`.

gaearon:
Awesome job.

chibicode:
@gaearon thanks!

Issue #92, opened 2015-09-01
Examples not working with version 2.0.0


khronnuz:
Hi, todomvc and counter examples are not working for me. I am getting:

```
ERROR in /Users/jhonny/dev/redux-devtools/src/createDevTools.js
Module not found: Error: Cannot resolve module 'react-redux/lib/components/createAll' in /Users/jhonny/dev/redux-devtools/src
 @ /Users/jhonny/dev/redux-devtools/src/createDevTools.js 19:40-87
```

mariopeixoto:
Same error here:

ERROR in /Users/mario/Documents/workspace/redux-devtools/src/createDevTools.js
Module not found: Error: Cannot resolve module 'react-redux/lib/components/createAll' in /Users/mario/Documents/workspace/redux-devtools/src
 @ /Users/mario/Documents/workspace/redux-devtools/src/createDevTools.js 19:40-87

ERROR in /Users/mario/Documents/workspace/redux-devtools/src/react/LogMonitorEntry.js
Module not found: Error: Cannot resolve module 'react-json-tree' in /Users/mario/Documents/workspace/redux-devtools/src/react
 @ /Users/mario/Documents/workspace/redux-devtools/src/react/LogMonitorEntry.js 19:21-47

ERROR in /Users/mario/Documents/workspace/redux-devtools/src/react/LogMonitorEntryAction.js
Module not found: Error: Cannot resolve module 'react-json-tree' in /Users/mario/Documents/workspace/redux-devtools/src/react
 @ /Users/mario/Documents/workspace/redux-devtools/src/react/LogMonitorEntryAction.js 21:21-47

mariopeixoto:
Looks like a 'npm install' in the root of the project is needed instead of only in the root of the example.

knowbody:
would be great to mention that in the example docs so ppl are not confused

PR #93, opened 2015-09-01
Improve instructions on how to run example

khronnuz:
This is related to #92 
Note: taken as it is from the root Readme

### DIFF
--- a/examples/counter/README.md
+++ b/examples/counter/README.md
@@ -1,6 +1,14 @@
 # Redux DevTools Counter example
 
-## Getting Started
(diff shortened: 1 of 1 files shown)


gaearon:
Thank you for sending a PR, very appreciated btw. :+1:

khronnuz:
Sure! Thanks for providing such exciting tools. Looking forward to be using them for my project and I will look for other opportunities to contribute to back (such as https://github.com/mjackson/expect/pull/20) ;)

PR #95, opened 2015-09-03
add link to Deku monitor

nivanson:


gaearon:
This seems like both a monitor + a Redux DevTools component re-implementation.
Maybe it's better to separate them?

nivanson:
Yeah. I can separate them. I can do a pull request to redux-devtools with a deku-folder. Does that make sense or do you prefer if it is maintained outside of your repo?

I was going to re-implement the entire implementation of your default monitor too but it turned out to be quite a lot since JSONTree had to be re-implemented too.

gaearon:
I'd rather prefer it in a separate repo, e.g. `redux-devtools-deku`.

nivanson:
Ok, thanks for the feedback. Will separate. Closing PR. Maybe opening a new one when done. :+1:

PR #99, opened 2015-09-10
Added missing thunk import.

alesch:


gaearon:
This would imply we depend on thunk middleware, but we don't.
I'll tweak the example to make it clearer.

PR #101, opened 2015-09-11
add actionDeserializer callback to persistState

conundrumer:
Because of some fancy middleware caching, I found that it was best if I passed Immutable objects through actions. They become plain objects after they get serialized in `persistState`, so I added a third argument to `persistState`, similar to #78

Example:

```es

function actionDeserializer(action) {
  if (action.type === 'SOME_ACTION_WITH_IMMUTABLE_PAYLOAD') {
    return {
      ...action,
      payload: Immutable.fromJS(action.payload)
    }
  }
  return action
}
persistState(sessionId, null, actionDeserializer)

```

This would also be useful for the folks who are using symbols instead of string constants for action types.

### DIFF
--- a/src/persistState.js
+++ b/src/persistState.js
@@ -1,20 +1,32 @@
-export default function persistState(sessionId, deserializer = null) {
+export default function persistState(sessionId, stateDeserializer = null, actionDeserializer = null) {
   if (!sessionId) {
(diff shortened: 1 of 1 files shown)


FredyC:
Look good :+1:

gaearon:
Out in 2.1.1.

Issue #100, opened 2015-09-10
Warning: DebugPanel(...): React component classes must extend React.Component.


billyjanitsch:
The latest version of React (v0.14-rc1) throws a warning for devtools components (DebugPanel, LogMonitor, LogMonitorEntry) because they [don't extend `React.Component`](https://github.com/gaearon/redux-devtools/blob/master/src/react/DebugPanel.js#L31).

tonyalaribe:
Yeah... I'm equally experiencing this.

gaearon:
Fixed by #103.

gaearon:
Out in 2.1.1.

PR #104, opened 2015-09-11
avoid react@0.14.0-rc1 warnings

quirinpa:
I've added extend React.Component to the classes that were missing it and removed references to findDomNode to avoid warnings with react@0.14.0-rc1.

### DIFF
--- a/src/react/DebugPanel.js
+++ b/src/react/DebugPanel.js
@@ -28,7 +28,7 @@ export function getDefaultStyle(props) {
   };
 }
 
-export default class DebugPanel {
+export default class DebugPanel extends React.Component {
   static propTypes = {
     left: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


FredyC:
Seems we have two identical PR :) #103 and also an issue #100. I kinda like this variant more, it's better to see `React.Component` instead of just `Component` in my opinion.

gaearon:
I merged #103 as I prefer destructuring for consistency.

I'm happy to accept `findDOMNode` change as a separate PR, but only after 0.14 ships, because we don't want to break people using 0.13 yet.

quirinpa:
Ofc i can totally understand @gaearon.

@FredyC thanks! I didn't realise the other PR :P

gaearon:

PR #107, opened 2015-09-13
use react-dock for debug panel docking

alexkuz:
Video: http://quick.as/v1zyudwqe

Caveats:

1. I had to move `DebugPanel` into `DevTools`, because it have a state now and needs to be connected to `redux`. Maybe provide legacy `DebugPanel` to client app that doesn't actually do anything?

2. `setMonitorState` is renamed to more appropriate (I suppose) `setPanelState` - again, should there be a legacy method? (or maybe shouldn't be renamed at all).

quickcast-13-09-2015-05-32-18

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "redux": "^2.0.0"
   },
   "dependencies": {
+    "react-dock": "0.0.2",
     "react-json-tree": "^0.1.8",
     "react-mixin": "^1.7.0",
     "react-redux": "^2.0.0",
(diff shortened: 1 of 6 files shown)


rockingskier:
Love the resizing!

gaearon:
I'm going to put this on hold until extracting LogMonitor into a separate project.

gaearon:
Closing, but I'd like to use `react-dock` this way—just without building its support into `devTools`.
Let's track this in https://github.com/gaearon/redux-devtools/pull/132 instead.

PR #110, opened 2015-09-14
React 0.14: Use of react-dom has moved findDOMNode

qnm:
As mentioned in [this post](https://facebook.github.io/react/blog/2015/07/03/react-v0.14-beta-1.html) `react-dom`has a number of methods originally part of `react`. One of these methods is `findDOMNode` which is used by `LogMonitor`.

This commit uses `findDOMNode` from `react-dom` as per advice issued in the deprecation warning.

### DIFF
--- a/src/react/LogMonitor.js
+++ b/src/react/LogMonitor.js
@@ -1,4 +1,5 @@
-import React, { PropTypes, findDOMNode, Component } from 'react';
+import React, { PropTypes, Component } from 'react';
+import { findDOMNode } from 'react-dom';
 import LogMonitorEntry from './LogMonitorEntry';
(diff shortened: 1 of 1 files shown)


igl:
And a little later they mention that you don't need to call findDOMNode on ref's at all: https://facebook.github.io/react/blog/2015/07/03/react-v0.14-beta-1.html#dom-node-refs

`const node = this.refs.elements` would be sufficient without loading react-dom at all.

This change could wait until `0.14.0-stable` is released

gaearon:
Fair. I'll close this out then. Happy to accept a PR removing `findDOMNode` call altogether after 0.14 is out.

gaearon:

PR #114, opened 2015-09-16
LogMonitor: remove calls to findDOMNode

igl:
`React.findDOMNode()` is deprecated and not a necessity anymore.
This patch should be merged and published after react-0.14.*-stable is out. (as discussed in #110)

While doing this i noticed that React is not mentioned as a dev or peerDependencies.
Is this intentional?

### DIFF
--- a/src/react/LogMonitor.js
+++ b/src/react/LogMonitor.js
@@ -1,4 +1,4 @@
-import React, { PropTypes, findDOMNode, Component } from 'react';
+import React, { PropTypes, Component } from 'react';
 import LogMonitorEntry from './LogMonitorEntry';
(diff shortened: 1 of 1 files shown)


itsmepetrov:
+1

nickhudkins:
+1

nikgraf:
+1

gaearon:
Thanks. This is fixed in https://github.com/gaearon/redux-devtools-log-monitor that is being split away from the main repo. I don't intend to merge this here, so closing. See https://github.com/gaearon/redux-devtools/pull/132 to track the split.

gaearon:

Issue #113, opened 2015-09-16
Implementing Undo/Redo collides with devtools


FredyC:
This is another case of actions that are being replayed with devtools which causes inconsistency in history used for undo/redo operation. Here is what I've invented so far...

    export function createHistory(limit = 100) {

        const recordedActions = [
            EditorActionType.SET_ROW_TEXT
        ];

        let currentIdx = 0;
        const states = [];
        const actions = [];

        function record(state, action) {
            // don't record if limit has been reached
            if (currentIdx >= limit) {
                return;
            }

            states[currentIdx] = state;
            actions[currentIdx] = action;
            currentIdx += 1;

            // clean forward history to forbid REDO upon changes
            states.splice(currentIdx);
            actions.splice(currentIdx);
        }

        function travel(state, distance = 1) {
            const targetIdx = currentIdx + distance;
            if (states[targetIdx] == null) {
                return state;
            }
            currentIdx = targetIdx;
            return states[currentIdx];
        }

        const reducer = (next, initialState) => (state, action) => {
            if (action.type === CommandActionType.UNDO_ACTION) {
                return travel(state, -(action.payload.distance));
            }
            if (action.type === CommandActionType.REDO_ACTION) {
                return travel(state, action.payload.distance);
            }
            if (recordedActions.includes(action.type)) {
                record(state, action);
            }
            return next(state, action);
        };

        return { reducer };
    }

I am then composing the `reducer` function together with application reducers.

Now if I dispatch some actions, it records all of them every time from beginning with new one added.  Also the `travel` function suffers even worst problem, all UNDO/REDO actions are replayed also causing to skip much further in history than wanted.

First issue can be avoided with check like `actions.includes(action)`, but it sounds bad enough to have a code just to cover devtools issue. However I don't see a way how to fix the second issue.

gaearon:
I already explained how to fix this: please contribute tests so I feel safe making optimizations and can remove the re-evaluation. https://github.com/gaearon/redux-devtools/issues/81#issuecomment-133209702

gaearon:
That said this reducer violates Redux constraints. It is not pure because it has local state (`states` and `actions`). That's why it doesn't work. If you move `state` and `actions` inside of its Redux state it will work as you'd like.

FredyC:
Alright, fair point. It might be actually easier to have each state to refer to previous state it can be undo-ed to.

gaearon:
Example unit test:

```js
import { createStore } from 'redux';
[...]

PR #116, opened 2015-09-17
tag recomputed actions

omnidan:
I'm writing a store enhancer and had lots of bugs when working with devtools and my own store enhancer. This is the only way I could prevent recomputed actions from the devtools to mess with my store enhancer.

I also tried including my store enhancer first and devtools first. Neither option worked fine.

I hope this is a good way to handle these kinds of issues, if not, please let me know a different way to do this.

### DIFF
--- a/src/devTools.js
+++ b/src/devTools.js
@@ -60,6 +60,8 @@ function recomputeStates(reducer, committedState, stagedActions, skippedActions)
 
   for (let i = 0; i < stagedActions.length; i++) {
     const action = stagedActions[i];
+    
+    if (i !== stagedActions.length - 1) action._recomputed = true;
 
     const previousEntry = computedStates[i - 1];
     const previousState = previousEntry ? previousEntry.state : committedState;


gaearon:
Can you provide a repro case for the kind of problems that you had?

omnidan:
@gaearon sorry it took a while, I was finishing up the [project](https://github.com/omnidan/redux-undo) - you can see how I'm using this tag here: https://github.com/omnidan/redux-undo/blob/master/src/index.js#L77-L80

You can try out the project here: https://github.com/omnidan/redux-undo-boilerplate to reproduce the error - Just download it and enable devtools in `webpack.config.js` by setting `__DEVTOOLS__` to `true`.

Without my fix, undo/redo gets really buggy, because actions are replayed by the devtools and there is no way to tell if an action is a new action or a recomputed one. It spams the history a lot and causes glitches when doing undo/redo a couple times.

Another solution would be tagging the new actions, but I feel like tagging the recomputed ones makes more sense, since they come from `redux-devtools`, if you rely on the `new` tag, redux would have to support it by default or you'll have to check if devtools are enabled etc

gaearon:
The issue is actually expected: `redux-undo` has local state (`store._history`) and violates Redux expectations. This is not how Undo/Redo should work in Redux. Please see https://github.com/gaearon/redux-devtools/issues/113#issuecomment-140792434 for a similar explanation.

gaearon:
Please see https://github.com/omnidan/redux-undo/issues/1 and https://github.com/omnidan/redux-undo/issues/2.

omnidan:
@gaearon thank you so much for your input. I'll try to adjust it asap.

omnidan:
This is fixed in [redux-undo](https://github.com/omnidan/redux-undo) now, thanks to @gaearon for all the suggestions / links / etc :ok_hand:

PR #122, opened 2015-09-20
fix(DevTools, LogMonitor): prevent LogMonitor to render once as visible state

nikgraf:
I noticed that LogMonitor is rendered once, because `isVisible` is set to true in the initial state.
see https://github.com/gaearon/redux-devtools/blob/master/src/devTools.js#L89

In the LogMonitor the `isVisible` state was updated, but that will only affect the second render cycle. This is an issue in case you try to create a universal/isomorphic application with the same template.

screen shot 2015-09-20 at 19 54 08

I propose to fix it by updating the props optimistically. As an alternative we could do this case inside the LogMonitor. Let me know what you think :smile:

### DIFF
--- a/src/createDevTools.js
+++ b/src/createDevTools.js
@@ -10,9 +10,26 @@ export default function createDevTools(React) {
     ActionCreators
   )
   class DevTools extends Component {
+
+    static defaultProps = {
+      visibleOnLoad: true
+    };
+
+    componentWillMount() {
+      const monitorState = {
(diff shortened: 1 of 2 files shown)


gaearon:
Thanks for the PR! I'm closing because the way monitors currently manage their state isn't composable.
Please track the solution to this in `next` branch: https://github.com/gaearon/redux-devtools/pull/132#issuecomment-143550058

PR #123, opened 2015-09-20
Add ChartMonitor as a LogMonitor companion

romseguy:
Hey, here's a draft for the implementation of [d3-state-visualizer](https://github.com/romseguy/d3-state-visualizer) as a monitor for Redux dev tools :-)

gifrecord_2015-09-20_221638

I disabled the tooltip you could see [here](https://github.com/romseguy/redux-store-visualizer) and exposed a `onClickText` hook from `TreeChart` API.

The idea would be to expand the corresponding node in LogMonitor JSON tree by clicking on the node text. Now I'm not really familiar with react-json-tree but maybe we can use [this](https://github.com/chibicode/react-json-tree/blob/master/src/mixins/expanded-state-handler.js) somehow?

I'm not sure whether we want this in a side panel as it's narrowing the app space too much. Maybe a different toggle key combination similar to LogMonitor's CTRL+H and set DebugPanel width/height>1?

Opinions and suggestions are welcome of course :+1:

### DIFF
--- a/src/react/ChartMonitor.js
+++ b/src/react/ChartMonitor.js
@@ -1,0 +1,131 @@
+import React, { PropTypes, Component } from 'react';
+import * as themes from './themes';
+import visualizer from 'd3-state-visualizer';
+
+const styles = {
+  container: {
+    fontFamily: 'monaco, Consolas, Lucida Console, monospace',
+    position: 'relative',
+    overflowY: 'hidden',
+    width: '100%',
+    height: '100%',
+    minWidth: 300
+  },
+  buttonBar: {
+    textAlign: 'center',
+    borderBottomWidth: 1,
+    borderBottomStyle: 'solid',
+    borderColor: 'transparent',
+    zIndex: 1,
+    display: 'flex',
+    flexDirection: 'row'
+  },
+  elements: {
+    position: 'absolute',
+    left: 0,
+    right: 0,
+    top: 38,
+    bottom: 0,
+    overflowX: 'hidden',
+    overflowY: 'auto'
+  }
+};
+
+export default class LogMonitor extends Component {
+  constructor(props) {
+    super(props);
+  }
+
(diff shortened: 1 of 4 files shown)


src/react/ChartMonitor.js
```diff
@@ -0,0 +1,131 @@
...
+    left: 0,
+    right: 0,
+    top: 38,
+    bottom: 0,
+    overflowX: 'hidden',
+    overflowY: 'auto'
+  }
+};
+
+export default class LogMonitor extends Component {
```
Shenlok: Wrong class name

gaearon:
This is really cool. Can I ask you to publish this as a separate project? We're going to extract `LogMonitor`, too.

subpopular:
This is ridiculously awesome. Wondering if we could keep collapsed nodes collapsed between state changes - seems the entire tree expands back out with every change. Would also be interested in how to handle overlapping nodes and paths. I'm going to dig in and play around with it myself in the meantime.

Also, I think this works best in it's own window, as seen [here](https://gist.github.com/tlrobinson/1e63d15d3e5f33410ef7).

Great work!

subpopular:
Was able to hook into the TreeChart's click method and use JSONTree to parse the objects on the right...

sep 21 2015 20 59

subpopular:
This works great as a standalone tool even without the LogMonitor. I put together an example [here](https://github.com/romseguy/d3-state-visualizer/issues/2).

romseguy:
@gaearon ok let's move here https://github.com/romseguy/redux-chart-monitor
@subpopular awesome suggestion, and it would be even better if you could move the issue to the above repo as well. It's more related to redux-devtools than d3-state-visualizer.

gaearon:
I'm closing but I'd appreciate you sending a PR to add it to README with the gif.

PR #132, opened 2015-09-24
New API

gaearon:
This is a work in progress in extracting `LogMonitor` into a [separate project](https://github.com/gaearon/redux-devtools-log-monitor) and simplifying the API.

The `todomvc` example is currently broken, but `counter` example works.
A few notable changes:

* `<LogMonitor>` is gone. It is now [in a separate repo](https://github.com/gaearon/redux-devtools-log-monitor).

* `<DevTools>` is gone. You are expected to use monitors directly and pass their arbitrary props *and* `store={store}` to them. The monitors are now expected to wrap their root component like this:

```js
import { connectMonitor } from 'redux-devtools';

class MyMonitor extends Component {
  /* business as usual */
}

export default connectMonitor(MyMonitor);
```

This lets the consumer use them directly:

```js
export default class Root extends Component {
  render() {
    const { store } = this.props;
    return (
      <div>
        <Provider store={store}>
          <CounterApp />
        </Provider>
        <Dock position='right' isVisible dimMode='none'>
          <LogMonitor store={store} />
        </Dock>
      </div>
    );
  }
}
```

**Please hold off updating custom monitors until we settle down on the API.**

* `<DebugPanel>` is no more. I want the recommended solution to support resizing, moving and collapsing the panel. [React Dock](https://www.npmjs.com/package/react-dock) looks promising, but I'm not sure how to hook it up to persist the state without [hardcoding this into the devTools logic](https://github.com/gaearon/redux-devtools/pull/107). Ideally “monitor decorators” (one of which could be based on `react-dock`) should be able to hold their state in Redux devTools store independently. I'm not sure how to handle this yet, so suggestions are welcome.


gaearon:
Any feedback on this from the monitor authors?

romseguy:
Looks fine and straight to the point. I'm on the verge of implementing an additional tree chart displaying the succession of actions leading to current state, and connected to the state tree chart as to make commit, toggle, and rollback features accessible by left/right click on chart nodes. I think it'd be great to have indexes and parent indexes on `stagedActions` so I could display a hierarchy of actions and allow the user to pick and choose between multiple action paths.

gaearon:
Oh, this is interesting. I'm probably going to push up another change here related to `monitorState`. I want monitors to (optionally) export a reducer to handle it—that would solve issues like #122 and give monitors more flexibility.

gaearon:
Okay, so I pushed this update which should let monitors be more powerful and potentially composable: https://github.com/gaearon/redux-devtools/commit/4ac31f985a13b66f291a6b96f97f75fb594daa59

The idea is that instead of `setMonitorState()`, the monitors now receive `monitorActions` which monitors themselves may choose by passing a parameter to `connectMonitor()`. The monitors may also specify their own reducer. When it exists, pass this reducer to `devTools()` as the first parameter, and it will manage `monitorState`.

Here is an example of how `LogMonitor` now manages its own visibility: https://github.com/gaearon/redux-devtools-log-monitor/commit/05312f0484a5c8f14686c5c683030d6f07d376e5. This should fix bugs like https://github.com/gaearon/redux-devtools/pull/122.

tomatau:
This proposed setup has invalid checksum problems on universal render because of react-dock and vendor prefixes. There's an issue here https://github.com/alexkuz/react-dock/issues/2 but it makes the redux-devtools currently not support React14rc in a universal build without any errors/warnings.

NB. This issue boils up to Radium who have been trying to solve the whole server side prefix issue for quite some time now.

gaearon:
Thanks for taking a look! That's work in progress, so no biggie. I'm trying to figure out the correct API for stateful monitors so it will be easy to use other docking solutions too. (We'll make sure whatever we include by default works fine with universal rendering.) Speaking of the right API, I'm onto something—going to change everything here again.

gaearon:
I just changed everything again but I'm closing in on the API I *really* like.
You can try the `counter` example now. (Never mind the failing tests: they are out of date.)

* Monitor composition! Monitors can contain other monitors, and each of monitors can have independent state that will be correctly serialized.
* Non-awkward API! You just nest components and enumerate monitor reducers in the reverse order.

As example, the [log monitor](https://github.com/gaearon/redux-devtools-log-monitor) now remembers the last scroll top (if you use `persistState` of course). The new `DockMonitor` (to be separated into another project) remembers its position (switch it with Ctrl+D) and whether it is visible (switch it with Ctrl+H). Both can accept options modifying their behavior in reducer configuration. And the API looks like this:

```js
const finalCreateStore = compose(
  applyMiddleware(
    thunk
  ),
  devTools(
    LogMonitor.createReducer({ // <---------------- new stuff
      preserveScrollTop: true
    }),
    DockMonitor.wrapReducer({ // <---------------- new stuff
      position: 'right',
      isVisible: true
    })
  ),
  persistState(
    window.location.href.match(
      /[?&]debug_session=([^&]+)\b/
    )
  )
)(createStore);
```

```js
export default class Root extends Component {
  render() {
    const { store } = this.props;
    return (
      <div>
        <Provider store={store}>
          <CounterApp />
        </Provider>
        <DevToolsProvider store={store}> // <---------------- new stuff
          <DockMonitor>  // <-------------------- new stuff
            <LogMonitor />
          </DockMonitor>
        </DevToolsProvider>
      </div>
    );
  }
```

Neat!

cc @lapanoid — it will be much easier to implement multi-docks and other uber-monitors

gaearon:
Some fun things you can do:

```js
//
// TODO: extract to a separate project.
//

import React, { Children, Component, PropTypes } from 'react';
import Dock from 'react-dock';
import MapProvider from './MapProvider';
import { connect } from 'react-redux';
import { combineReducers } from 'redux';

function mapUpstreamStateToDownstreamState(state, i) {
  return {
    devToolsState: state.devToolsState,
    monitorState: state.monitorState[i]
  };
}

function wrapReducers(reducers) {
  return combineReducers(reducers);
}

@connect(state => state)
export default class CombineMonitors extends Component {
  static propTypes = {
    monitorState: PropTypes.object
  };

  render() {
    const { position, isVisible, children } = this.props;
    return (
      <div>
        {Children.map(children, (child, i) =>
          <MapProvider mapState={(state) => mapUpstreamStateToDownstreamState(state, i)}>
            {child}
          </MapProvider>
        )}
      </div>
    );
  }
}

CombineMonitors.wrapReducers = wrapReducers;
```

```js
  devTools(
    CombineMonitors.wrapReducers([
      DockMonitor.wrapReducer({ position: 'right' })(
        LogMonitor.createReducer({})
      ),
      DockMonitor.wrapReducer({ position: 'left' })(
        LogMonitor.createReducer({})
      )
    ])
  ),
```

```js
export default class Root extends Component {
  render() {
    const { store } = this.props;
    return (
      <div>
        <Provider store={store}>
          <CounterApp />
        </Provider>
        <DevToolsProvider store={store}>
          <CombineMonitors>
            <DockMonitor>
              <LogMonitor />
            </DockMonitor>
            <DockMonitor>
              <LogMonitor />
            </DockMonitor>
          </CombineMonitors>
        </DevToolsProvider>
      </div>
    );
  }
}
```

screen shot 2015-09-28 at 05 03 50

examples/counter/src/dock/DockMonitor.js
```diff
@@ -0,0 +1,111 @@
...
+    }
+    e.preventDefault();
+
+    const key = event.keyCode || event.which;
+    const char = String.fromCharCode(key);
+    switch (char) {
+    case 'H':
+      this.props.toggleVisibility();
+      break;
+    case 'D':
```
SimenB: Any chance of making these customizable?
I'm asking because `ctrl+d` is add bookmark in every browser I know (not when using Mac, but Linux/Windows).

gaearon: Yeah, sure. It's just proof of concept now. :-)
I was focusing on getting monitors to compose & persist local state together—not exact props API.

lapanoid:
@gaearon Cool! Thanks for the hint. I will peer to new api.

eapache:
> Please hold off updating custom monitors until we settle down on the API.

@gaearon I am playing with writing a react-native monitor for this, is it worth continuing against the current API or should I just hold off until this PR is ready? cc @boourns

gaearon:
@eapache The API in this PR is settled, I think.

gaearon:
Redux DevTools 3.0.0 Beta 2 is out.
Please try it: https://github.com/gaearon/redux-devtools/releases/tag/v3.0.0-beta-2.

dariocravero:
Looking good :)! Didn't have any problems to update by following what was done on the examples on `next`.

chentsulin:
It's may be worth to tell people should install `redux-devtools-log-monitor` and `redux-devtools-dock-monitor` with `@1.0.0-beta-2`;

```
npm i --save-dev redux-devtools@3.0.0-beta-2 redux-devtools-log-monitor@1.0.0-beta-2 redux-devtools-dock-monitor@1.0.0-beta-2
```

IwanKaramazow:
Mac OS X El Capitan 10.11 here. 
The key-bindings seem to work just fine on safari (9.0 (11601.1.56)) and Chrome (46.0.2490.71 (64-bit)). 
On Firefox 41.0.2 I get:
```'ReferenceError: event is not defined'  ```
whenever trying to change the location/hide Redux Devtools. 

Is this a known issue? 
Seems like a page loads enormously slow on firefox (instant on chrome/safari) and at some point webpack_hmr gets disconnected.

**Comment by gaearon** [comment]
It's not a known issue. Where does the error occur?

**Comment by IwanKaramazow** [comment]
The extremely slow loading had something to do with my webpack config: devtool #eval-source-map...
Switching back to eval-devtool gives me normal load time. 


The ReferenceError: event is not defined traces back to;
```
DockMonitor.prototype.handleKeyDown = function handleKeyDown(e) {
	    if (!e.ctrlKey) {
	      return;
	    }
	    e.preventDefault();

	    var key = event.keyCode || event.which;  //==> firefox highlights this line of code.
	    var char = String.fromCharCode(key);
	    switch (char.toUpperCase()) {
	      case this.props.toggleVisibilityKey.toUpperCase():
	        this.props.dispatch(_actions.toggleVisibility());
	        break;
	      case this.props.changePositionKey.toUpperCase():
	        this.props.dispatch(_actions.changePosition());
	        break;
	      default:
	        break;
	    }
	  };
```
Did you try the latest version of DevTools on Firefox?

**Comment by gaearon** [comment]
Whoops, that `event` reference was totally not intended, thanks for catching this.

**Comment by gaearon** [comment]
Fixed in https://github.com/gaearon/redux-devtools-dock-monitor/commit/ecc67af0e3f973d0bd915453253e6b1f88c1407c, this will get into the next release.

**Comment by IwanKaramazow** [comment]
Glad I could help :)

**Comment by PeterKow** [comment]
Works great, thanks! 
If someone is interested in upgrade commit check this out -> https://github.com/PeterKow/redux-react-babel-webpack/commit/eff5418b19fbddaf7216f65d91ca207b8ea81783

**Comment by SimenB** [comment]
Not sure if I should report it here or in the redux-devtools-dock-monitor repo, but here goes :smile: 

This line https://github.com/gaearon/redux-devtools-dock-monitor/blob/ecc67af0e3f973d0bd915453253e6b1f88c1407c/src/DockMonitor.js#L51 stop all ctrl-shortcuts working, including ctrl+r, which is annoying. Could you only prevent default if it's actually handled in one of the `case`s?

**Comment by gaearon** [comment]
Thanks for reporting, indeed that repo is the correct place.

**Comment by chentsulin** [comment]
@gaearon Will 3.0 have a helper for create DevTools on new window, or just do something like this?

```js
// containers/Devtools
import React from 'react';
import { createDevTools } from 'redux-devtools';
import LogMonitor from 'redux-devtools-log-monitor';

export default createDevTools(
  <LogMonitor />
);

// utils/createDevToolsWindow.js
import React from 'react';
import { render } from 'react-dom';
import { Provider } from 'react-redux';
import DevTools from '../containers/DevTools';

/**
 * Puts Redux DevTools into a separate window.
 * Based on https://gist.github.com/tlrobinson/1e63d15d3e5f33410ef7#gistcomment-1560218.
 */
export default function createDevToolsWindow(store) {
  // Window name.
  const name = 'Redux DevTools';

  // Give it a name so it reuses the same window.
  const win = window.open(
    null,
    name,
    'menubar=no,location=no,resizable=yes,scrollbars=no,status=no'
  );

  // Reload in case it's reusing the same window with the old content.
  win.location.reload();

  // Set visible Window title.
  win.document.title = name;

  // Wait a little bit for it to reload, then render.
  setTimeout(() => render(
    <Provider store={store}>
      <DevTools />
    </Provider>,
    win.document.body.appendChild(document.createElement('div'))
  ), 100);
}
```

chentsulin:
Or maybe a SideEffect Compoent `<DevToolsWindow>`?

kumar303:
I just tried hooking up dev tools for the first time with the latest beta (3) and everything is pretty much working fine. The only thing for me is the key bindings don't do anything. No errors, just no hiding when I press `H` or `h`. It didn't work in any browser I tried so maybe I [hooked it up wrong](https://github.com/kumar303/music-arranger/blob/master/public/js/lib/components/dev-tools.js)? This was [my patch](https://github.com/kumar303/music-arranger/commit/f1b9c22302f95f606b1dac6b1d710ff1496fc3e2) to add the dev tools.

SimenB:
It's `ctrl-h` and `ctrl-q`

kumar303:
Oh, duh, thanks! That works in Chrome. I still get the ReferenceError in FF but it looks like that fix hasn't been released yet.

Hobby-Student:
Hey out there,

hope this one is the right topic to report. I'm from Germany and so the "@" is used by "AltGr+q" (the same as "Ctrl+Alt+q"), which is prevented by the DockMonitor... (using FF for testing)

This is the original code:
```
DockMonitor.prototype.handleKeyDown = function handleKeyDown(e) {
    if (!e.ctrlKey) {
      return;
    }
    e.preventDefault();

    var key = e.keyCode || e.which;
    var char = String.fromCharCode(key);
    switch (char.toUpperCase()) {
      case this.props.toggleVisibilityKey.toUpperCase():
        this.props.dispatch(_actions.toggleVisibility());
        break;
      case this.props.changePositionKey.toUpperCase():
        this.props.dispatch(_actions.changePosition());
        break;
      default:
        break;
    }
  };
```
which I modified a bit:
```
DockMonitor.prototype.handleKeyDown = function handleKeyDown(e) {
    if (!e.ctrlKey || e.altKey) {                               //<--- this one
      return;
    }
    e.preventDefault();

    var key = e.keyCode || e.which;
    var char = String.fromCharCode(key);
    switch (char.toUpperCase()) {
      case this.props.toggleVisibilityKey.toUpperCase():
        this.props.dispatch(_actions.toggleVisibility());
        break;
      case this.props.changePositionKey.toUpperCase():
        this.props.dispatch(_actions.changePosition());
        break;
      default:
        break;
    }
  };
```

SimenB:
@Hobby-Student gaearon/redux-devtools-dock-monitor#2

Hobby-Student:
@SimenB thanks... and sry for hijacking this one ;)

Issue #125, opened 2015-09-21
Documentation and sample code mismatch


gouegd:
Hey there! :)

The main readme says:

> DevTools is a store enhancer, which should be added to your middleware stack after applyMiddleware as applyMiddleware is potentially asynchronous

And just after:

> ```
const finalCreateStore = compose(
  // Enables your middleware:
  applyMiddleware(m1, m2, m3), // any Redux middleware, e.g. redux-thunk
  // Provides support for DevTools:
  devTools(),
  // Lets you write ?debug_session=<name> in address bar to persist debug sessions
  persistState(window.location.href.match(/[?&]debug_session=([^&]+)\b/))
)(createStore);
```

However, `compose`d functions are called right to left, so in this code `devTools()` will be called _before_ `applyMiddleware`, not _after_ like the doc says we must.

gaearon:
You're right about the call order.
The intention was to encourage user to *write* it after (= below) `devTools()` in the `compose` chain.

If you see a better way of phrasing this please send a PR.

PR #136, opened 2015-09-29
Rephrasing how to order the devtools and middleware calls

gouegd:
As discussed in this issue: https://github.com/gaearon/redux-devtools/issues/125

This attempts to clarify how `devTools` should be called before `applyMiddleware` but may appear before in the code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ A live-editing time travel environment for [Redux](https://github.com/rackt/redu
 npm install --save-dev redux-devtools
 ```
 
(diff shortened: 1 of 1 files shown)


gaearon:
Hmm. I still think most people will read “before” and put it before `applyMiddleware()` inside `compose()`.

gouegd:
Right... I made a second, simpler attempt at this.

gaearon:
:+1: Thanks!

Issue #139, opened 2015-10-02
Create UMD build


ohana54:
Hi,

Continuing the discussion in #130 about AMD support here.

You (@gaearon) mentioned in #130 that you don't plan to provide an AMD version of Redux DevTools, just like there is none for redux itself.
I probably didn't explain myself correctly - I'm not looking for an "AMD" version, I just need it to be UMD compliant.
Redux is built using webpack so I can use it (and currently do) with RequireJS.
Redux DevTools is not built using webpack so I can't use it (well, I tried and failed).
I hope this explanation is better.
I'm of course willing to submit a PR that tries to add support, I just want to make sure I'm on the right track here.

Thanks for your help and patience :)

gaearon:
Oh, I see. We can probably add a UMD build. Can you take a look at how `react-redux` repo does it with Webpack config and submit a PR doing it in a similar fashion?

ohana54:
Sure, that was my plan :)
I'll try to make it happen during the weekend.

ohana54:
Created #140

gaearon:
Hey, I'm really sorry for the churn but I split the packages already.
Can you please make separate PRs to:

* this project (against the `next` branch)
* https://github.com/gaearon/redux-devtools-log-monitor
* https://github.com/gaearon/redux-devtools-dock-monitor

Sorry if that's too much, but it's hard to move forward when everything is in one place. :-(

ohana54:
No problem, I'll take care of it

PR #140, opened 2015-10-03
Create UMD Build

ohana54:
I have 2 notes about this change:
* I separated the build to 2 packages: the logical parts and the react components. If there's a better way to put everything in one file please let me know.
* I used "connect" instead of "createAll" from react-redux, it was easier this way to regard react-redux as an external library. I didn't know how to do it otherwise, so again, if there's a better way let me know.

I tested the result with my RequireJS project and it works great.
I also tested the regular build (from lib folder) to make sure that the modified react-redux import didn't break anything.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 .DS_Store
 lib
 coverage
+dist
--- a/package.json
+++ b/package.json
@@ -5,7 +5,12 @@
   "main": "lib/index.js",
   "scripts": {
     "clean": "rimraf lib",
-    "build": "babel src --out-dir lib",
+    "build:lib": "babel src --out-dir lib",
+    "build:umd": "webpack src/index.js dist/redux-devtools.js --config webpack.config.development.js",
+    "build:umd:min": "webpack src/index.js dist/redux-devtools.min.js --config webpack.config.production.js",
+    "build:umd:react": "webpack src/react/index.js dist/redux-devtools.react.js --config webpack.config.development.js",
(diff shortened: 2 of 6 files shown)


gaearon:
Hi, thanks! Is it essential that I merge this now, or can you wait until next major release?

ohana54:
It's not urgent at all

SimenB:
@gaearon Hey, I was wondering why this (and thus redux itself) has the `screw-ie8`-option for uglify? React supports IE8, shouldn't redux as well?

We use them both from npm, so no issue for us, but people using the prebuilt versions might run into obscure issues, right?

gaearon:
@SimenB To tell the truth I didn't intend to support IE8—that it even works speaks to the relatively few code Redux has, not to a conscious effort. I have no means of testing it in IE8 anyway, so I don't try.

SimenB:
You can download IE8 images directly from Microsoft here: https://dev.modern.ie/tools/vms

I don't expect fancy things like the devtools to work, but explicitly munging up the code so it might _not_ work seems weird to me when Facebook has stated that at least version 1.0 of React is going to support IE8

gaearon:
I'm not Facebook (at least not yet :wink: ). If somebody has a requirement of supporting IE8, I welcome that somebody to review the project, set up a testing infrastructure to avoid regressions, and review pull requests for IE8 compatibility. Just “not mangling” isn't enough, and I don't want to give the wrong impression that IE8 is officially supported. Use it at your own risk.

gaearon:
I'm closing this particular PR because https://github.com/gaearon/redux-devtools/issues/139#issuecomment-148908749.

PR #144, opened 2015-10-07
:arrow_up: react-json-tree 0.2.0

chibicode:
this version should support react 0.14.


gaearon:
I'll close as I did this in https://github.com/gaearon/redux-devtools-log-monitor.

PR #150, opened 2015-10-12
Bump react-json-tree dependency to 0.2.0

pbomb:
Version 0.2.0 support React 0.13.1 and 0.13.2 instead of requiring at least 0.13.3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "redux": "^2.0.0 || ^3.0.0"
   },
   "dependencies": {
-    "react-json-tree": "^0.1.9",
+    "react-json-tree": "^0.2.0",
     "react-mixin": "^1.7.0",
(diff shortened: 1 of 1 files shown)


chentsulin:
duplicate of #144

gaearon:
We'll require 0.14 as minimal React version soon anyway.

pbomb:
That's unfortunate for me. I will get our app upgraded to 0.13.3 so we can use the devtools but it's going be awhile before we can move to 0.14 (due to some dependencies we have), so we'll probably be stuck on this version for some time.

gaearon:
Please try 3.0 Beta 2 and leave feedback.
It includes React 0.14 support: https://github.com/gaearon/redux-devtools/releases/tag/v3.0.0-beta-2

Issue #164, opened 2015-10-21
findDOMNode - use on react-dom instead of react


ondrejbartas:
React.findDOMNode is deprecated. Please use ReactDOM.findDOMNode from require('react-dom') instead.

ondrejbartas:
#165

gaearon:
From README:

<img width="577" alt="screen shot 2015-10-21 at 16 28 16" src="https://cloud.githubusercontent.com/assets/810438/10637901/c36a41da-7810-11e5-82cc-b2f8eb5722e4.png">

Also, using search will give you:

<img width="856" alt="screen shot 2015-10-21 at 16 28 54" src="https://cloud.githubusercontent.com/assets/810438/10637917/de2f1496-7810-11e5-9630-87de0892ed47.png">

The warnings are gone in [3.0 beta](https://github.com/gaearon/redux-devtools/releases/tag/v3.0.0-beta-2). Please use it. We will not fix them in 2.0 because it's a breaking change anyway.

Issue #152, opened 2015-10-12
[Announcement] Please don't create React 0.14-related PRs


gaearon:
Hi all,

Thanks a lot for sending 0.14-related PRs.
I know it's out, and I'm working on releasing updates for Redux-related packages.

Currently, sending PRs doesn't help because it's not hard for me to fix the warnings—the issue is in coordinated release of all packages (React Redux, DevTools, React Transform, etc) and fixing compatibility issues in the tooling. I will do my best to transition everything to 0.14 this week.

**In the meantime, please don't send PRs related to React 0.14 to Redux and related projects.**

Thanks.

nickhudkins:
If there's anything any of us can do to help out please let us know :)

gaearon:
Please try 3.0 Beta 2 and leave feedback.
It includes React 0.14 support: https://github.com/gaearon/redux-devtools/releases/tag/v3.0.0-beta-2

gaearon:
Reopening because people keep creating them :stuck_out_tongue: 
Use 3.0 Beta!

PR #165, opened 2015-10-21
Fix warning for findDOMNode

ondrejbartas:


gaearon:
https://github.com/gaearon/redux-devtools/issues/152
https://github.com/gaearon/redux-devtools/issues/164#issuecomment-149894434

PR #168, opened 2015-10-23
Use react-dom.findDOMNode instead of react.findDOMNode

leeight:
Thanks


gaearon:
Use 3.0 Beta: https://github.com/gaearon/redux-devtools/releases/tag/v3.0.0-beta-2
See also https://github.com/gaearon/redux-devtools/issues/152.

leeight:
OK

PR #173, opened 2015-10-28
WIP - Add LinkToState

rileylark:
Ahoy! This is a PR with an idea we use at Pear Deck for fast collaboration between designers & developers. The idea is that once you have the state in a single object, you might as well render it to the URL! Once your states are in urls, you can share them around your development and design teams. I've ported it over to the redux devtools. 

[image](https://drive.google.com/file/d/0Bz-K4z5jMfRzYk5FZmRMWXplU2s/view?pli=1)

There are some rough edges and no tests, but I figured I'd shoot this over and see what you think before continuing. Let me know if I should adhere to a different format or anything!

### DIFF
--- a/examples/counter/containers/App.js
+++ b/examples/counter/containers/App.js
@@ -1,16 +1,20 @@
 import React, { Component } from 'react';
 import CounterApp from './CounterApp';
 import { createStore, applyMiddleware, combineReducers, compose } from 'redux';
-import { devTools, persistState } from 'redux-devtools';
+import { devTools, persistState, urlState } from 'redux-devtools';
 import { DevTools, DebugPanel, LogMonitor } from 'redux-devtools/lib/react';
 import thunk from 'redux-thunk';
 import { Provider } from 'react-redux';
 import * as reducers from '../reducers';
 
+const devStateJsonMatches = window.location.href.match(/[?&]dev_state=([^&]+)\b/);
+const devStateJson = devStateJsonMatches ? devStateJsonMatches[1] : null;
+
 const finalCreateStore = compose(
   applyMiddleware(thunk),
   devTools(),
-  persistState(window.location.href.match(/[?&]debug_session=([^&]+)\b/))
+  persistState(window.location.href.match(/[?&]debug_session=([^&]+)\b/)),
+  urlState(devStateJson),
 )(createStore);
 
 const reducer = combineReducers(reducers);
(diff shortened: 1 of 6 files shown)


rileylark:
Ah, I am just understanding about the `next` branch and also #132 . 

Let me know if this basic idea is something you'd like to include by default @gaearon and I can update this PR to point wherever you think is best. I get the feeling that `master` is not best atm ;)

rileylark:
After working with this some more, we've noticed that the middleware approach I started down in this PR is overkill - you really just need a component with access to the entire state and some code to set the initial state at `createStore` time. Then you don't need any middleware.

But: having such a link in the devtools by default is pretty awesome, and might help win more people over to the idea of redux. Maybe it's worth finding a way to include the link by default?

@gaearon if you like the idea of including this state link by default I can think about it some more. If not we can separate it into an easy-to-plug module.

gaearon:
My suggestion is to change this to react to a *key press* and use `prompt()` instead of a dedicated UI both for importing and exporting state.

calesce:
@gaearon: I've implemented something similar to this approach, however with ```prompt()``` there is a character limit in Chrome of about 2000, which isn't great (see [here](http://stackoverflow.com/questions/18088662/text-passed-to-window-prompt-gets-replaced-with)).

rileylark:
The ux of copying and pasting is much more cumbersome on the user (the
devs) so I've decided to stick with the ui in this PR in my own projects.
However, I found a much simpler way to code it, without a store enhancer at
all: you only need to parse the initial state when... Setting the initial
state!

If the ui here becomes interesting to you Dan, I'd be happy to submit an
updated PR. The prompt approach isn't very interesting to me but I don't
mind if someone else wants to pick this up and go that direction.

Thanks again for a great set of packages!

rileylark:
My OCD needs me to close this PR. Feel free to use any of my code in this PR with or without attribution, if you like!

gaearon:
Hey, sorry, I literally have zero time maintaining this project right now 😞 

With the current architecture I think this could work better as a separate custom monitor. Then `DockMonitor` lets you switch between multiple monitors with a keyboard shortcut, so you can switch between `LogMonitor` and `LinkToStateMonitor` with a single key. I hope this helps!

rileylark:
Oh, no problem at all. Thanks for the response!

PR #175, opened 2015-10-29
Fix getting persist session id from url

peteruithoven:


gaearon:
Using your code in new README—thanks!

PR #194, opened 2015-11-28
Rethrow errors as console.error looses sourcemaps

### LINKED ISSUE #193: Source maps not working
Hi there,

I use [este dev stack](https://github.com/este/este) and when I appended a redux-devtools, then source maps stopped working - when some error occure it does not show the exact file where it occures...

zalmoxisus:
Fix #193.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -67,7 +67,7 @@ function computeNextEntry(reducer, action, state, error) {
     nextState = reducer(state, action);
   } catch (err) {
     nextError = err.toString();
-    console.error(err.stack || err);
+    setTimeout(() => { throw err; }, 0);
   }
 
   return {


gaearon:
Can you remind me some context with "before" and "after" this change?
How does it solve the problem, and does it work great in all browsers?

zalmoxisus:
As described [here](https://github.com/webpack/webpack/issues/1646), `console.error` and `console.trace` just miss the source maps. As a result we aren't able to see the the line or the stack lines which caused the error (we see just `eval`s). According to [the chrome issue](https://code.google.com/p/chromium/issues/detail?id=376409), it wouldn't work as V8 knows nothing about source maps. So, we have to rethrow the error. Also, a setTimeout workaround helps to execute it in a different process, not to halt the current function.

zalmoxisus:
@gaearon, actually, you're right, such issues should be dealt by Webpack and browsers vendors. And it seems to work now in Chrome 47, though not so accurate, but it's better than using workarounds.

Just so you know, BEFORE this patch applied:
screen shot 2015-12-20 at 11 50 42 am

AFTER:
screen shot 2015-12-20 at 11 52 14 am

But at least we can now track errors. In Chrome 46 there was only the bundle with evals.

gaearon:
I don't really mind merging so pleased reopen if it works. I just wanted to remember what it solves so thank you for the screenshots.

zalmoxisus:
Since none of these implementations solves anything in Safari and `console.error(err.stack)` [works just fine in Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=583083), I had better submit another pull request with this workaround only for Chrome.

PR #202, opened 2015-12-12
Update readme for v3 release

ellbee:
I just tried to fix up stuff that I thought might block the release.

* Updated Examples to v3
* Added a bit about the root.dev and root.prod pattern
* Removed examples that don't have v3 versions yet. My idea was to ping the maintainers and then re-add them to the readme as they get updated. Not sure if this is the right approach, but I was concerned that people might get confused by the differing APIs.

What did I miss/get wrong?


gaearon:
Thanks a lot! I released 3.0

PR #205, opened 2015-12-15
Fix typo in README.md

cesarandreu:
The word was `quckly` => `quickly`, since the diff makes it a bit difficult to see.


ellbee:
Thanks!

PR #208, opened 2015-12-20
Solve the issue with loosing sourceMaps in Chrome

### LINKED ISSUE #193: Source maps not working

zalmoxisus:
Fix #193. Related to #194.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -67,7 +67,10 @@ function computeNextEntry(reducer, action, state, error) {
     nextState = reducer(state, action);
   } catch (err) {
     nextError = err.toString();
-    console.error(err.stack || err);
+    if (typeof window === 'object' && typeof window.chrome !== 'undefined') {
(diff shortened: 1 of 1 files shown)


gaearon:
Is there any benefit to hardcoding a chrome check here? I'd rather have a single way of doing thing as long as this method doesn't bring any downsides in other browsers.

zalmoxisus:
Yes, unfortunately, I figured that it has downsides in Firefox. When clicking to the error line in the console, Firefox cannot open the file from sourceMaps. Though it works well with `console.error(stack)`.
As about Safari, it shows only the error tree without any lines in both cases.

gaearon:
Got it, thanks.

gaearon:
Out in 3.0.1.

PR #211, opened 2015-12-26
Warn and do not attempt to render if store not instrumented

### LINKED ISSUE #179: Add a warning for missing instrument() call
In the `next` branch, we should warn if the user forgets to add `DevTools.instrument()` to the `createStore` composition chain.

[This](https://github.com/gaearon/redux-devtools/blob/1752cccd82319f5c30c069423035b09ffb7b3bf0/src/createDevTools.js#L32) is a good place to do that. Something like

```js
if (!this.liftedStore) {
  console.error(
    'Did you forget to call DevTools.instrument() when creating store?'
  );
}
```

robertknight:
Very confusing error messages were produced if the
container component returned by createDevTools()
was constructed without a store being injected or
if the store had not been instrumented with
DevTools.instrument()

Add warnings for both of these cases and avoid console error
spam by short circuiting rendering of the devtools if they
have not been setup correctly.

Fixes #179

### DIFF
--- a/src/createDevTools.js
+++ b/src/createDevTools.js
@@ -24,14 +24,32 @@ export default function createDevTools(children) {
 
     constructor(props, context) {
       super(props, context);
+
+      if (!props.store && !context.store) {
+        console.error(
+`DevTools cannot render because no Redux store was injected via \
+props or <Provide>`);
+        return;
+      }
+
       if (context.store) {
         this.liftedStore = context.store.liftedStore;
       } else {
         this.liftedStore = props.store.liftedStore;
       }
+
+      if (!this.liftedStore) {
+        console.error(
+`DevTools cannot render because the store not been instrumented \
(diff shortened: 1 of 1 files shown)


src/createDevTools.js
```diff
@@ -24,14 +24,32 @@ export default function createDevTools(children) {
 
     constructor(props, context) {
       super(props, context);
+
+      if (!props.store && !context.store) {
+        console.error(
+`DevTools cannot render because no Redux store was injected via \
+props or <Provide>`);
```
gaearon: Typo: should be `<Provider>`.

gaearon:
I'll get this in and fix the message. Thanks a bunch! :+1:

gaearon:
Out in 3.0.1.

robertknight:
Thanks for the quick review & merge :+1:

Issue #213, opened 2015-12-29
Expected identifier error in IE8


kevana:
Same as [react-redux #232](https://github.com/rackt/react-redux/issues/232), also affects dock-monitor and log-monitor.

kevana:
If the solution is to switch back to CommonJS modules for now I can put a PR together

gaearon:
If you have time please do create the PRs. Note that for backward compat the monitors should still have default exports, so the entry point might look pretty much like the code generated by Babel.

gaearon:
We decided to not support IE8. React is also dropping IE8 compat guarantees in the next version.

PR #214, opened 2015-12-30
Go back to CommonJS modules for now...

kevana:
...until https://phabricator.babeljs.io/T2817 is resolved.

First crack at the conversion, not entirely sure if this works but I'll verify tomorrow.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -1,6 +1,7 @@
-import difference from 'lodash/array/difference';
+const difference = require('lodash/array/difference');
 
-export const ActionTypes = {
+
+const ActionTypes = {
   PERFORM_ACTION: 'PERFORM_ACTION',
   RESET: 'RESET',
   ROLLBACK: 'ROLLBACK',
@@ -14,7 +15,7 @@ export const ActionTypes = {
 /**
  * Action creators to change the History state.
  */
-export const ActionCreators = {
+const ActionCreators = {
   performAction(action) {
     return { type: ActionTypes.PERFORM_ACTION, action, timestamp: Date.now() };
   },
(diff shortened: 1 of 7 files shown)


src/instrument.js
```diff
@@ -342,4 +343,7 @@ export default function instrument(monitorReducer = () => null) {
     const liftedStore = createStore(liftReducer(reducer));
     return unliftStore(liftedStore, liftReducer);
   };
-}
+};
+
+module.exports.ActionTypes = ActionTypes;
```
kevana: Is this kosher for achieving a default and named exports?

kevana:
Fixes #213

SimenB:
@gaearon Could this be looked at? :)

@kevana Tests doesn't need to change, do they?

kevana:
@SimenB I don't think it needs any test changes, module behavior should be the same as before. I did add exports.spec.js as a sanity check for exports. I dropped this version into my apps and didn't see any issues but I'm still not completely confident that I got the default and named exports correct.

gaearon:
Thanks, but I think I won't support DevTools on IE8.
Even Microsoft [ends support for IE < 11 tomorrow](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support).
You can use something like `es3ify` as a global transformation on your project if you need it.

kevana:
I understand completely, it's an easy fix with webpack. Hopefully the really long tail of IE8 traffic starts dying a bit faster...

SimenB:
IE8 will still be supported in enterprise, so my project is stuck with it for now...

In this case we can just use es3ify though, so no real problem. As long as no getters are used, I'm happy :smile:

gaearon:
Microsoft is not going to support it for anybody, not even enterprise customers. If some enterprises still decide to use unsupported version without security fixes at this point it's their problem.

kevana:
The security fixes are their problem but the lost revenue is my problem. It's hard for me to make a business case for dropping support when a material percentage of our customers still use it. Hopefully this pushes the stragglers to finally upgrade.

gaearon:
It's a pain for library authors and it's gotta go at some point. I understand your concerns but JS libraries are going this way now that MS doesn't support it, and at this point the only solution we can offer is transpile to ES3 by yourself or use older versions of the tools. Cheers.

PR #222, opened 2016-01-09
Fix webpack typo: providePlugin -> definePlugin

joshwcomeau:
A couple of places mention the wrong webpack plugin required for excluding DevTools from production builds. I spent a very confusing morning trying to figure out why providePlugin wasn't working =)

I also expanded the explanation a bit for webpack users, since it wasn't at all intuitive to me that you could use non-plain variables for the key (eg. `process.env.NODE_ENV`, and not just a normal variable like `NODE_ENV`), or that you had to stringify the value (`JSON.stringify('production')`).

This is my first open-source PR. Thanks for all the amazing work with Redux and Redux Dev Tools!

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,14 +144,30 @@ export default function configureStore(initialState) {
 
 #### Exclude DevTools from Production Builds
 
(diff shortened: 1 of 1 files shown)


gaearon:
:+1:

PR #220, opened 2016-01-08
Missing semicolon throwing TodoMVC error

yepnamesjames:
Sorry if this one is too wild. It is my first open-source PR.

PR #227, opened 2016-01-18
Fix class property in todomvc example

mheiber:
Fix error caused by missing semicolon in the class property declaration.

Relevant part of the ES proposal: https://github.com/jeffmo/es-class-fields-and-static-properties#145-class-definitions

Full error message:

> ERROR in ./components/Footer.js
Module build failed: SyntaxError: redux-devtools/examples/todomvc/components/Footer.js: A semicolon is required after a class property (18:3)


yepnamesjames:
#220

gaearon:
Fixed by #220, thanks though!

PR #231, opened 2016-01-22
Added docs on Reset / Revert / Sweep / Commit buttons

wesbos:
I didn't understand what these buttons did until I played around with them for a few minutes - I thought some nice documentation on what they do would be needed.


gaearon:
Thanks! They are described here: https://github.com/gaearon/redux-devtools-log-monitor#features
Maybe we can just link to the log monitor README more prominently?

wesbos:
Ah - I never saw that. Makes sense - I'll close this.

PR #232, opened 2016-01-25
Add error on undefined action type

skovhus:
First, thanks for a great development tool! 

I like the helpful errors in Redux, e.g. for undefined `action.type`. Of course you should have unit test catching this, but it would be nice if `redux-devtools` also had this message in case the action type is `undefined`. 

I ported the behaviour from here:
https://github.com/rackt/redux/blob/5b94a18f560eff60ce37ae01d1b16387bcfb8e3a/src/createStore.js#L110

Let me know what you think. : )

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -16,6 +16,12 @@ export const ActionTypes = {
  */
 export const ActionCreators = {
   performAction(action) {
+    if (typeof action.type === 'undefined') {
+      throw new Error(
+        'Actions may not have an undefined "type" property. ' +
+        'Have you misspelled a constant?'
+      );
+    }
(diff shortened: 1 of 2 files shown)


gaearon:
Wouldn't Redux throw this anyway? Is logging it just before throwing beneficial?

skovhus:
No, this is not caught by Redux as all the actions are wrapped.

gaearon:
Oh, this is a great catch. I think it should be a hard error, then.

skovhus:
Yes, it is somewhat hidden that devtools wraps the action and hides these errors. : )

Ok, I will make it a hard error.

skovhus:
@gaearon please have a look now.

PR #235, opened 2016-01-30
Check instrumentation store enhancer's reducer and not to be included twice

zalmoxisus:
Related to https://github.com/zalmoxisus/redux-devtools-extension/issues/36


zalmoxisus:
Added another error related to https://github.com/zalmoxisus/redux-devtools-extension/issues/47.

sbuzonas:
Don't suppose there's a way to have more than one live in harmony. Trying to merge two apps into one both with custom monitors that only make sense to a particular context.

Using two docks seemed like the easy approach, but perhaps it's easier to combine them instead?

PR #240, opened 2016-02-04
Show error message, not just a stack trace

sapegin:
Before:

After:

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -77,7 +77,7 @@ function computeNextEntry(reducer, action, state, error) {
       // In Chrome, rethrowing provides better source map support
       setTimeout(() => { throw err; });
     } else {
-      console.error(err.stack || err);
+      console.error(err);
     }
   }
 
--- a/test/instrument.spec.js
+++ b/test/instrument.spec.js
@@ -169,7 +169,7 @@ describe('instrument', () => {
     expect(computedStates[3].error).toMatch(
       /Interrupted by an error up the chain/
     );
-    expect(spy.calls[0].arguments[0]).toMatch(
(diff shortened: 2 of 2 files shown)


zalmoxisus:
I don't think duplicating errors is a good way to fix it. Better to just replace console.error(err.stack || err)` with `console.error(err)`, so we'll get the message along with the stack.

I guess `err.stack` was added to fix the problem with Chrome, and since it is now fixed there by rethrowing the error, we don't need it anymore.

zalmoxisus:
With `console.error(err.stack)` we get
screen shot 2016-02-04 at 5 10 02 pm
With `console.error(err)`
screen shot 2016-02-04 at 5 09 50 pm

sapegin:
Cool, I’ve updated the PR.

gaearon:
Can you please check before/after in Chrome, Firefox, and Safari?

zalmoxisus:
I've tested as well in Safari. It doesn't affect Chrome. But we also have to change tests as we get the error object not a string now.

So,
```js
expect(spy.calls[0].arguments[0]).toMatch(
      /ReferenceError/
    );
```
should become
```js
expect(spy.calls[0].arguments[0].toString()).toMatch(
      /ReferenceError/
    );
```

gaearon:
:+1:

sapegin:
Tests are fixed.

sapegin:
@zalmoxisus @gaearon What can I do to make this merged?

gaearon:
Thanks for reminding. I will push out a patch later today.

gaearon:
Out in 3.1.1.

PR #241, opened 2016-02-04
add maxAge option

echenley:
- adds `options` parameter to instrument. Only `maxAge` property is used currently but could conceivably be used for other options.
- removes oldest actions once maxAge is reached
- should take care of [memory issues](https://github.com/gaearon/redux-devtools-log-monitor/issues/17) caused by high-frequency actions

Usage:

```js
DevTools.instrument({ maxAge: 30 });
```

### DIFF
--- a/src/createDevTools.js
+++ b/src/createDevTools.js
@@ -7,9 +7,6 @@ export default function createDevTools(children) {
   const monitorProps = monitorElement.props;
   const Monitor = monitorElement.type;
   const ConnectedMonitor = connect(state => state)(Monitor);
-  const enhancer = instrument((state, action) =>
-    Monitor.update(monitorProps, state, action)
-  );
 
   return class DevTools extends Component {
     static contextTypes = {
@@ -20,7 +17,10 @@ export default function createDevTools(children) {
       store: PropTypes.object
     };
 
-    static instrument = () => enhancer;
+    static instrument = (options) => instrument(
+      (state, action) => Monitor.update(monitorProps, state, action),
+      options
+    );
 
     constructor(props, context) {
       super(props, context);
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -172,7 +172,7 @@ function liftReducerWith(reducer, initialCommittedState, monitorReducer, options
 
       for (let i = 0; i < idsToDelete.length; i++) {
         if (computedStates[i + 1].error) {
-          // Stop if error is found. Commit only up to error.
+          // Stop if error is found. Commit actions up to error.
           excess = i;
           idsToDelete = stagedActionIds.slice(1, excess + 1);
           break;
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -235,15 +235,18 @@ function liftReducerWith(reducer, initialCommittedState, monitorReducer, options
         break;
       }
       case ActionTypes.PERFORM_ACTION: {
-        if (options.maxAge && stagedActionIds.length === options.maxAge) {
+        if (
+          options.maxAge &&
+          stagedActionIds.length === options.maxAge &&
+          !computedStates[1].error
+        ) {
           // If maxAge has been reached, auto-commit earliest non-@@INIT action.
           delete actionsById[stagedActionIds[1]];
           skippedActionIds = skippedActionIds.filter(id => id !== stagedActionIds[1]);
           stagedActionIds = [0].concat(stagedActionIds.slice(2));
           committedState = computedStates[1].state;
           computedStates = computedStates.slice(1);
-        }
-        if (currentStateIndex === stagedActionIds.length - 1) {
+        } else if (currentStateIndex === stagedActionIds.length - 1) {
           currentStateIndex++;
         }
         const actionId = nextActionId++;
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -238,7 +238,7 @@ function liftReducerWith(reducer, initialCommittedState, monitorReducer, options
         if (options.maxAge && stagedActionIds.length === options.maxAge) {
           // If maxAge has been reached, remove oldest action.
           delete actionsById[stagedActionIds[1]];
-          skippedActionIds = skippedActionIds.filter(id => id !== stagedActionIds[0]);
+          skippedActionIds = skippedActionIds.filter(id => id !== stagedActionIds[1]);
           stagedActionIds = [0].concat(stagedActionIds.slice(2));
           committedState = computedStates[1].state;
           computedStates = computedStates.slice(1);
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -165,12 +165,21 @@ function liftReducerWith(reducer, initialCommittedState, monitorReducer, options
       computedStates
     } = liftedState;
 
-    function commitExcessActions() {
+    function commitExcessActions(excess) {
       // If maxAge has been exceeded, auto-commit excess.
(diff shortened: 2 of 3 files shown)


gaearon:
Thank you for your work on this! I’ll take a look very soon.

zalmoxisus:
@echenley, wouldn't it delete the INIT action? I guess we shouldn't do it.

echenley:
@zalmoxisus 

1. What's the point of holding onto the init action? RESET, COMMIT, and ROLLBACK all recreate it:

```js
actionsById = { 0: liftAction(INIT_ACTION) };
```

2. I could remove [line 343](https://github.com/gaearon/redux-devtools/pull/241/files#diff-9dd65be703999bb051f530e79908a0c9R243) to keep a reference to the last committed state even when it is removed from the list. Currently it's essentially auto-committing each time it overflows, which might not be so intuitive. Which behavior would users be more likely to expect?

gaearon:
I think the first action should always be `INIT` so effectively you want to squash all “too old” actions into a single `INIT` action.

zalmoxisus:
Actually, I was going to implement the same approach for chrome extension some time ago, but came out with auto-commit for the same reason.

We need auto-commiting in the extension as we serialize and send the lifted state object, and don't want it to grow too much. Here we don't have this problem, the performance issue comes only when rendering the monitor.

So, I think better to implement it as a composable monitor. In this way we keep lifted state as is, but in the log monitor we get only the last actions, and we can change the maximum allowed actions prop or remove it on the fly (as HMR takes care of it).

You may take [redux-devtools-filter-actions](https://github.com/zalmoxisus/redux-devtools-filter-actions) as an example or extend it as there we also delete actions and states and it is also a kind of filtering.

gaearon:
Interesting. I agree it’s interesting to try doing this as a monitor and compare performance on projects where large amounts of actions accumulate very quickly. (http://erikras.github.io/redux-form is a good example, cc @erikras)

gaearon:
However, I still think a “rolling log” like this that doesn’t grow memory forever is useful and I’d love to see it as an option here. The PR just needs a little more work to squash the old actions into a single INIT action.

zalmoxisus:
@gaearon, I tested this [here](http://zalmoxisus.github.io/redux-devtools-extension/examples/router/) (though for the extension), even added there redux-router as it adds extra weight with circular dependencies. Most all the CPU was taken by the app's rendering.

zalmoxisus:
If we keep it here, wouldn't be better to pass this limit as a prop to the created `<DevTools/>` as we pass the `store` (when not using context)? So the HMR could take care of it when changing the limit value on the fly? I may be mistaken and HMR updates also when changing store enhancer, but I don't think so.

echenley:
What about adding an optional property to the commit action, either an index, or a number of actions to commit which defaults to all? That way any monitor can commit actions explicitly. I'll fiddle with it and see how it feels.

Thinking something like:

```js
case ActionTypes.COMMIT: {
  if (liftedAction.index) {
    // commit only up to liftedAction.index
    // keep rest of history
  } else {
    // commit everything
  }
}
```

echenley:
@gaearon I updated the code to always keep `@@INIT` at the top and just auto-commit the action after it.

@zalmoxisus It does seem like HMR doesn't work with this, though it's probably not something that would be changed much. I'd be happy to attempt a different approach in another PR.

zalmoxisus:
@echenley, thanks for investigating this! After merging, I'll try to get it work with the extension to be able to change the value on the fly.

gaearon:
1. Set `maxAge` to 5
2. Make an error in reducer
3. Dispatch 10 actions

Actual: the first 5 actions are skipped because autocommit happened during the error state. Expected: I don’t think actions are ever expected to disappear. I think that auto-committing should only go as far as the last non-erroring action. If a reducer threw, that action should not be auto-committed until it is fixed. However once it is fixed, auto-commit should take action immediately if we are over `maxAge`.

gaearon:
Also, it seems like we are not taking care of `currentStateIndex`. It is used by monitors like sliders. We want to make sure that if it is earlier than the auto-committed action, that we move it to be *at least* the minimal action index that still exists.

gaearon:
Finally, we want to run more tests. A few tests for this specific functionality does not seem enough. We might want to add a copy for every “normal” test that uses `maxAge` and shows how it affects the expected behavior. If we did that, we would not have “suprises” I mentioned above.

echenley:
Thanks for the feedback! Making progress on this. The issue with `currentStateIndex` should be fixed now. I'll be adding more tests in the next couple days and cleaning up a bit, but the functionality is there.

Here's a gif of the error behavior. I have `maxAge` set to 3. It doesn't commit any actions once it hits an error, but will auto-commit them when the error is fixed and the reducer is reloaded.

devtools

echenley:
I think the tests are fairly complete now. In `instrument.js`, I extracted the auto-commit logic into a function since it has to run when performing actions and when reloading the reducer. This also make it pretty trivial to allow the commit action to accept either an index or # of actions to commit as previously mentioned (i.e. just call that function with the number or actions to commit).

amay:
How's this looking for release? I have an app with hi-frequency actions and this change is exactly what I'm looking for. I tested this branch out and I'm happy to report it fixes my problem. Now I just need a release to point to. Thanks for the good work!

amay:
@echenley any update on merging this?

echenley:
@amay I haven't heard anything. I imagine @gaearon is busy with his bazillion other repos. :wink:

Sinistralis:
Just to note, this solves issues with having an onChange handler tied to the store, which causes the filter monitor to lag the application like crazy after a few seconds of typing normally. Filter solves this as well, but I would prefer both knowing that there is a performance hit after a certain number of messages. It's throwing off testing slightly.

gaearon:
That’s awesome, thank you for hard work. :+1:

gaearon:
@echenley I added you as a collaborator. Please feel free to edit the documentation to include information about `maxAge`!

echenley:
Excellent, thank you!

mess0615:
I don't know if this is related, but lately I keep getting a lot of maxAge output in the console such as:

maxAge 50 50

Is this leftover debugging code?

echenley:
@mess0615 This PR didn't introduce any debug statements. Not sure where it's coming from, but it's not this repo. https://github.com/gaearon/redux-devtools/search?utf8=%E2%9C%93&q=console

rnsloan:
@mess0615 looks like it is coming from the Chrome extension https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd

kbsymanz:
I am getting these maxAge messages too. Just started today with no apparent relationship to code that I have changed. The messages stop when I disable Redux Dev Tools in Chrome.

Found this code in the Chrome tools that seems related.

```
            case g.PERFORM_ACTION:
                console.log("maxAge", r.maxAge, f.length),
                r.maxAge && f.length === r.maxAge && u(1),
                _ === f.length - 1 && _++;
                var M = d++;
                p[M] = c,
                f = [].concat(f, [M]),
                E = f.length - 1;
                break;
```

rnsloan:
Chrome extension issue https://github.com/zalmoxisus/redux-devtools-extension/issues/89

zalmoxisus:
Oh, sorry, just forgot to remove the console log. It is not related to this PR in any way.

PR #243, opened 2016-02-09
Export liftAction, liftReducerWith, and unliftState from instrument.js

MikeRyan52:
Related to ngrx/store#31, exporting these three functions lets us use all of redux-devtools' instrumentation code.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -131,14 +131,14 @@ function recomputeStates(
 /**
  * Lifts an app's action into an action on the lifted store.
  */
-function liftAction(action) {
+export function liftAction(action) {
   return ActionCreators.performAction(action);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


MikeRyan52:
I am going to go ahead and close this. While it certainly would have been awesome to share some underlying code, it has become apparent that instrumentation for @ngrx/store will end up looking pretty different over the coming months.

[Our fork lives here](https://github.com/ngrx/devtools)

zalmoxisus:
Having the same instrumentation for `ngrx/store` would give the ability to [use there may awesome monitors built for Redux DevTools via extension](https://github.com/zalmoxisus/redux-devtools-extension/issues/77). Though I think it would be better to [have the instrumentation separately from `createDevTools` not to depend on React](https://github.com/gaearon/redux-devtools/issues/245).

PR #248, opened 2016-02-13
Consecutive toggle

khankuan:
Added SET_ACTIONS_ACTIVE action for ability to set multiple actions between start and end indices to be active or skipped. This is needed to allow log monitor to toggle multiple actions at once via `shift+click`.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -1,4 +1,5 @@
 import difference from 'lodash/difference';
+import union from 'lodash/union';
 
 export const ActionTypes = {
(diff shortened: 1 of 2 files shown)


gaearon:
This looks reasonable! Sorry for a long time to reply. Any specific monitor where you intend to use this?

khankuan:
It's intended for the LogMonitor. Will work on that soon :)

gaearon:
:+1:

khankuan:
Hmm, im thinking of using state in LogMonitor, any opinion on that?

PR #251, opened 2016-02-20
fix regexp to extract debug session name

ganmor:
Extracting session name did not work when using url containing a hash

 ( For example when using hash history )

See #https://github.com/zalmoxisus/redux-devtools-extension/pull/59


gaearon:
Can you also do the same in the examples?

ganmor:
Sure. I have updated the PR

Issue #163, opened 2015-10-20
No more visibleOnLoad in v3?


erikras:
I'm using `redux-devtools@3.0.0-beta-3`, and I see that the `visibleOnLoad` prop has gone away.

Is there a new way to have it default to hidden?

erikras:
Nevermind. Figured it out. Using `DockMonitor`:
```js
<DockMonitor
    toggleVisibilityKey="H"
    changePositionKey="Q"
    defaultIsVisible={false}> // <--------------- Bingo!
  <LogMonitor/>
</DockMonitor>
```

gaearon:
That's it!

PR #260, opened 2016-03-11
Add defaultIsVisible to walkthrough docs

rickhanlonii:
## Overview

This PR adds documentation for hiding the DockMonitor on page load.

## Notes

- The README previously contained documentation on the `visibleOnLoad` property under "features" (see [here](https://github.com/gaearon/redux-devtools/pull/75/files)). But that prop and documentation has been removed. 

- While looking for a replacement, I found  `defaultIsVisible` in https://github.com/gaearon/redux-devtools/issues/163 and confirmed that it's still in use.

- I don't think this belongs under features in the README as before. Instead I added it to the usage section of the walkthrough, where it's likely to have the highest visibility.

### DIFF
--- a/docs/Walkthrough.md
+++ b/docs/Walkthrough.md
@@ -45,8 +45,10 @@ const DevTools = createDevTools(
   // Monitors are individually adjustable with props.
   // Consult their repositories to learn about those props.
   // Here, we put LogMonitor inside a DockMonitor.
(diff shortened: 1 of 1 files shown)


gaearon:
Thanks!

rickhanlonii:
No sweat!

PR #263, opened 2016-04-01
Provide information that action is being replayed

tomkis1:
There are obviously some libraries which are trying to reduce side effects in Reducers (https://github.com/salsita/redux-side-effects, https://github.com/raisemarketplace/redux-loop), however we need to maintain purity of these Reducers - side effect execution is always deferred. This is easy to implement for classic Reducers but we would need to make it work for devtools time travel as well.

There is a workaround and it's basically the way redux-loop is doing that by [overriding `dispatch`](https://github.com/raisemarketplace/redux-loop/blob/master/modules/install.js#L38-L42) drawback of this solution is though that modified `dispatch` is not used for redux init action and this is for example problem for reproducing Elm architecture in Redux because in Elm architecture there might be side effects in the init function.

That's why we [re-wrote redux-side-effects](https://github.com/salsita/redux-side-effects/commit/875ddc6be25977619cf76d7d6b17bbbac98f6aa9#diff-b2037c95647bbd93515ca58460ccc012L75), however this introduces issue with timetravel because we don't have any way to find out when the action is being replayed and when not. [This comment](https://github.com/salsita/redux-side-effects/issues/14#issuecomment-204150958) describes it quite accurately.

This PR does solve the problem, we need to decide about proper way to implement it though because right now `replaying` flag is passed as third argument to reducer. The question is though, is that really sane? Can't it break some functionality of other people? Maybe it would be better to provide the information directly in the action itself.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -59,7 +59,7 @@ const INIT_ACTION = { type: '@@INIT' };
 /**
  * Computes the next entry in the log by applying an action.
  */
-function computeNextEntry(reducer, action, state, error) {
+function computeNextEntry(reducer, action, state, error, replaying) {
   if (error) {
     return {
       state,
@@ -70,7 +70,7 @@ function computeNextEntry(reducer, action, state, error) {
   let nextState = state;
   let nextError;
   try {
-    nextState = reducer(state, action);
+    nextState = reducer(state, action, replaying);
   } catch (err) {
     nextError = err.toString();
     if (typeof window === 'object' && typeof window.chrome !== 'undefined') {
@@ -97,7 +97,8 @@ function recomputeStates(
   committedState,
   actionsById,
   stagedActionIds,
-  skippedActionIds
+  skippedActionIds,
+  replaying
 ) {
   // Optimization: exit early and return the same reference
   // if we know nothing could have changed.
@@ -120,7 +121,7 @@ function recomputeStates(
     const shouldSkip = skippedActionIds.indexOf(actionId) > -1;
     const entry = shouldSkip ?
       previousEntry :
-      computeNextEntry(reducer, action, previousState, previousError);
+      computeNextEntry(reducer, action, previousState, previousError, replaying);
 
     nextComputedStates.push(entry);
   }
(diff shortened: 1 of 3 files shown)


tomkis1:
As we had the discussion [via twitter](https://twitter.com/tomkisw/status/715910049634328576), it looks like passing the `replaying` flag as third argument may break functionality for people, who are using this argument for something else.

`redux-devtools` instrumentation relies on fact that non-standard actions like thunks for example have already been processed and converted to plain old objects. We can seize the fact and pack the `replaying` flag directly into the dispatched Action.

Last two commits fixes this.

src/instrument.js
```diff
@@ -70,7 +70,7 @@ function computeNextEntry(reducer, action, state, error, replaying) {
   let nextState = state;
   let nextError;
   try {
-    nextState = reducer(state, action, replaying);
+    nextState = reducer(state, { ...action, replaying });
```
tomkis1: I am in favour of keeping the original reference untouched, of course it is a performance penalty.

src/instrument.js
```diff
@@ -70,7 +70,7 @@ function computeNextEntry(reducer, action, state, error) {
   let nextState = state;
   let nextError;
   try {
-    nextState = reducer(state, action);
+    nextState = reducer(state, { ...action, replaying });
```
gaearon: What if the user already has a meaning for `replaying` field? Can this be a Symbol instead?

tomkis1: I've been thinking about serializability. However, it **does** make sense to ignore that field when actions are serialized anyway.

Agreed that the naming clash can potentially be an issue, will re-work this.

tomkis1: @gaearon Only issue I am seeing here is that `IsReplaying` Symbol would have to be exposed by `redux-devtools`. However, that would mean that any library which relies on the flag, would need to define `redux-devtools` as `peerDependency`.

Not sure if that's acceptable.

gaearon:
Ugh. I kinda feel bad about this. Maybe we should just fix initial dispatch in Redux?
Like we could dispatch lazily the first time `getState()` is accessed. :hushed:

gaearon:
If you’re up for it, can you look into fixing it in Redux itself? Even if it’s a breaking change. Then we can look at our options again and consider whether it’s worth it.

gaearon:
Please see https://github.com/reactjs/redux/issues/465.
Let’s discuss there instead.

tomkis1:
I thought it was feature not bug that middlewares are not applied for initial dispatch. If that's not the case then I absolutely agree with you!

Thanks, will look into the redux issue now.

PR #266, opened 2016-04-11
Support React >= 15 peer dependency

cchamberlain:
This semver range is breaking https://github.com/whetstone/redux-devtools-diff-monitor/pull/28. Can we get a React 15 version cut?

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "peerDependencies": {
     "redux": "^3.0.0",
-    "react": "^0.14.0 || ^15.0.0-rc.1"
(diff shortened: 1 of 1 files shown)


gaearon:
No, `^15.0.0-rc.1` matches `15.0.0` and any further stable 15.x versions. You can verify this on http://semver.npmjs.com/.

Please see an explanation for what’s likely to be your issue here: https://github.com/reactjs/react-redux/pull/342#issuecomment-207647767

cchamberlain:
Ahh my bad, the npm error had me confused in the travis build log but I can see that the rc matching range makes more sense. Thanks for explanation!

PR #267, opened 2016-04-13
maxAge should throw error when < 2

quicksnap:
`maxAge` with 1 breaks DevTools. Also, due to the @INIT action, it doesn't make sense to use anything < 2.

### DIFF
--- a/src/instrument.js
+++ b/src/instrument.js
@@ -411,6 +411,15 @@ function unliftStore(liftedStore, liftReducer) {
  * Redux instrumentation store enhancer.
  */
 export default function instrument(monitorReducer = () => null, options = {}) {
+  /* eslint-disable no-eq-null */
+  if (options.maxAge != null && options.maxAge < 2) {
+  /* eslint-enable */
+    throw new Error(
+      'DevTools.instrument({ maxAge }) option, if specified, ' +
+      'may not be less than 2.'
+    );
+  }
+
   return createStore => (reducer, initialState, enhancer) => {
 
     function liftReducer(r) {
--- a/test/instrument.spec.js
+++ b/test/instrument.spec.js
@@ -479,6 +479,12 @@ describe('instrument', () => {
 
       spy.restore();
     });
+
+    it('should throw error when maxAge < 2', () => {
+      expect(() => {
+        createStore(counter, instrument(undefined, { maxAge: 1 }));
+      }).toThrow(/may not be less than 2/);
+    });
   });
 
   describe('Import State', () => {


zalmoxisus:
On Chrome extension we use `0` value for disabling maxAge, which [fit the condition](https://github.com/gaearon/redux-devtools/blob/master/src/instrument.js#L286), but, yes, having `1` will make the history tree to be empty and we should throw an error. What do you think of prohibiting only `1`?

src/instrument.js
```diff
@@ -411,6 +411,10 @@ function unliftStore(liftedStore, liftReducer) {
  * Redux instrumentation store enhancer.
  */
 export default function instrument(monitorReducer = () => null, options = {}) {
+  if (options.maxAge && options.maxAge < 2) {
+    throw new Error('DevTools maxAge option may not be < 2.');
```
gaearon: Let’s change it to `DevTools.instrument({ maxAge }) option, if specified, may not be less than 2.`

src/instrument.js
```diff
@@ -411,6 +411,10 @@ function unliftStore(liftedStore, liftReducer) {
  * Redux instrumentation store enhancer.
  */
 export default function instrument(monitorReducer = () => null, options = {}) {
+  if (options.maxAge && options.maxAge < 2) {
```
gaearon: Let’s change it to `if (options.maxAge != null && options.maxAge < 2)` to also warn for `0`. To disable the option, they can pass `null` or `undefined`.

test/instrument.spec.js
```diff
@@ -479,6 +479,13 @@ describe('instrument', () => {
 
       spy.restore();
     });
+
+
+    it.only('should throw error when maxAge < 2', () => {
```
gaearon: `it.only`

quicksnap: One of these days I'm going to write an eslint rule for this..

quicksnap:
@gaearon Should we allow for zero for @zalmoxisus's use case?

quicksnap:
Whoops, I see you commented about zero _after_ his comment. I'll adjust the PR soon.

quicksnap:
Amended PR to log error when `maxAge: 0`, as throwing an error is a breaking change. Look ok, or should we throw for 0 as well?

Also fixed other feedback.

gaearon:
I think throwing is fine. It’s technically a breaking change but this is a development tool and 0 was never really a valid value.

quicksnap:
Amended again, just throws, and removed the warning.

src/instrument.js
```diff
@@ -411,6 +411,13 @@ function unliftStore(liftedStore, liftReducer) {
  * Redux instrumentation store enhancer.
  */
 export default function instrument(monitorReducer = () => null, options = {}) {
+  if (options.maxAge !== null && options.maxAge < 2) {
```
gaearon: I’d make it `!=` to make it clear `undefined` falls into the same bucket. Even though `<` would be false anyway. If a lint rule prevents this we should surround this with disable/enable comments.

test/instrument.spec.js
```diff
@@ -479,6 +479,13 @@ describe('instrument', () => {
 
       spy.restore();
     });
+
+
```
gaearon: Nit: extra newlines.

quicksnap:
I went with multi-line `eslint-disable`, but could change it to `// eslint-disable-line` if you prefer.

Fixed whitespace nit

gaearon:
Thanks. I’ll release this when I have some other fixes to do as it is very minor. Cheers!

zalmoxisus:
Agree. I was just lazy to add an extra checkbox
screen shot 2016-04-13 at 10 23 22 pm

But it's not a problem to convert `0` values to undefined to support already stored settings.

PR #275, opened 2016-04-27
Fix redux support for observables

### LINKED ISSUE #274: Devtools seem to break reduxs observable support
Redux implemented support for observables since v3.5.0. When I attach an observable to the store, I will receive the devtools state instead of my application state.

A quick example:

```javascript
import {Observable} from 'rxjs';
import {createStore, compose, applyMiddleware} from 'redux';
import thunk from 'redux-thunk';

const store = createStore(rootReducer, compose(
    applyMiddleware(thunk),
    window["devToolsExtension"] ? window["devToolsExtension"]() : f => f
));

Observable.from(store).map(state => console.log(state));
```

Will log something like `Object {monitorState: Object, actionsById: Object, nextActionId: 1, stagedActionIds: Array[1], skippedActionIds: Array[0]…}` when the store updates.

When I disable the redux devtools in Chrome, I get the expected state instance.

micha149:
This change fixes the support for observables when store is instrumented by devtools. This will resolve  #274. To be honest, I am not sure if there is not a more elegant solution instead of reimplementing the subscribe function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,15 +54,17 @@
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
     "rimraf": "^2.3.4",
+    "rxjs": "^5.0.0-beta.6",
     "webpack": "^1.11.0"
   },
   "peerDependencies": {
-    "redux": "^3.0.0",
+    "redux": "^3.5.0",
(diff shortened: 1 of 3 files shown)


gaearon:
Thanks! Should be out in `3.3.0`.

PR #279, opened 2016-05-04
clear cached "actionsById" on ActionTypes.COMMIT

amirlazarovich:
Committing creates a clean slate, however `initialLiftedState.actionsById` is left with old actions leaving an unnecessary memory footprint


zalmoxisus:
That's a good catch, but this problem will also occur for `RESET`, `ROLLBACK `, `IMPORT_STATE` and any other action that affects `actionsById`. It's happening because [we're mutating this object](https://github.com/gaearon/redux-devtools/blob/master/src/instrument.js#L294-L296), but this shouldn't affect `initialLiftedState`.

What do you think of the solution I've implemented in https://github.com/zalmoxisus/redux-devtools-instrument/pull/2/commits/48478e14de81af87c2fddf2d04024c5814f0bac8?

PR #305, opened 2016-07-18
Remove AppContainer from production

rajeshaz09:
1) updated react-hot-loader to 3.0.0-beta2 and moved it to
devDependencies
2) AppContainer included in development only.


rajeshaz09:
wrong commits

PR #306, opened 2016-07-18
Remove AppContainer from production

rajeshaz09:
1) updated react-hot-loader to 3.0.0-beta2 and moved it to
devDependencies
2) AppContainer included in development only.


yannuil:
I thought that you can leave `AppContainer` in the production code, as [it does nothing.](https://github.com/gaearon/react-hot-loader/blob/d360e3a24914db0c123ca03192bfdade48453751/src/AppContainer.prod.js)

PR #313, opened 2016-09-05
Browser extensions is available for Firefox too

lorenzos:
I almost started the manual integration process before noticing the mentioned Chrome extension is available for Firefox too.


zalmoxisus:
Thanks!

PR #322, opened 2016-10-27
Fixed bug 244, Npm install fails on windows

Durairaj:
Updated package.json to include cross-env in order to fix the bug 244


zalmoxisus:
Thanks a lot for addressing this!

PR #334, opened 2017-01-12
Allow react-redux 5 in peerDependencies

evocateur:
This allows users to upgrade to the API-compatible v5 of react-redux without encumbering their `node_modules` with a nested copy of react-redux (v4) under `redux-devtools/node_modules`.

It also moves redux and react-redux into `devDependencies`, which is the proper location for packages that are also specified in `peerDependencies`. Otherwise, peers specified as an alternating range (`||`) won't resolve properly. This brings it in line with what [react-redux does](https://github.com/reactjs/react-redux/blob/master/package.json#L94-L107).

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,17 +54,18 @@
     "react": "^0.14.0",
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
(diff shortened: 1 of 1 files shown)


zalmoxisus:
Thanks for the contribution!

PR #358, opened 2017-04-18
fixed PropTypes warnings

tsmirnov:
Removed Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "mocha": "^2.2.5",
     "mocha-jsdom": "^1.0.0",
     "react": "^0.14.9",
-    "prop-types": "^15.5.0",
+    "prop-types": "^15.5.7",
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
     "react-redux": "^4.0.0",
@@ -62,11 +62,12 @@
   },
   "peerDependencies": {
     "react": "^0.14.9 || ^15.3.0",
-    "prop-types": "^15.5.0",
+    "prop-types": "^15.5.7",
     "react-redux": "^4.0.0 || ^5.0.0",
     "redux": "^3.5.2"
   },
   "dependencies": {
+    "prop-types": "^15.5.7",
     "lodash": "^4.2.0",
     "redux-devtools-instrument": "^1.0.1"
   }
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "jsdom": "^6.5.1",
     "mocha": "^2.2.5",
     "mocha-jsdom": "^1.0.0",
-    "react": "^0.14.0",
+    "react": "^0.14.9",
     "prop-types": "^15.5.0",
     "react-addons-test-utils": "^0.14.0",
     "react-dom": "^0.14.0",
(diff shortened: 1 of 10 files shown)


csillag:
What is keeping this from being merged?

asotog:
any updates on this ?

package.json
```diff
@@ -52,6 +52,7 @@
     "mocha": "^2.2.5",
     "mocha-jsdom": "^1.0.0",
     "react": "^0.14.0",
+    "prop-types": "^15.5.0",
```
zalmoxisus: 1. Shouldn't it be in `dependencies` instead?

2. It's not compatible with `"react": "^0.14.0 || ^15.0.0"`, but only with `"react": "^0.14.9 || ^15.3.0"`, so peerDependencies should be modified accordingly.

3. Examples are meant to be independent (with their dependencies), so we should add it there as well.

package.json
```diff
@@ -61,7 +61,8 @@
     "webpack": "^1.11.0"
   },
   "peerDependencies": {
-    "react": "^0.14.0 || ^15.0.0",
+    "react": "^0.14.9 || ^15.3.0",
+    "prop-types": "^15.5.0",
```
zalmoxisus: I think it would be better to move it in `dependencies` as I  suggested before and as it's [recommended by the author](https://github.com/reactjs/prop-types#how-to-depend-on-this-package) (so one could not use that lib). It should be removed from `devDependencies`as well.

package.json
tsmirnov: 1. I think, there is no need to put it to dependencies if react is not there.
2. Nice catch! Thanks :)
3. Not sure if I understand what you mean by that.

package.json
tsmirnov: Yeah... you are right. Will fix it now.

package.json
zalmoxisus: 1. One could not use `prop-types` in his project. It should't prevent DevTools to work. React is mandatory because otherwise there will be no monitors.

3. Every example has it's own [`package.json`](https://github.com/gaearon/redux-devtools/blob/master/examples/counter/package.json) and we need dependencies to be present there as well.

package.json
```diff
@@ -62,11 +62,12 @@
   },
   "peerDependencies": {
     "react": "^0.14.9 || ^15.3.0",
-    "prop-types": "^15.5.0",
+    "prop-types": "^15.5.7",
     "react-redux": "^4.0.0 || ^5.0.0",
     "redux": "^3.5.2"
   },
   "dependencies": {
+    "prop-types": "^15.5.7",
```
zalmoxisus: If it's already in `dependencies` it will be installed anyway, so no need for having it also in `peerDependencies` and `devDependencies`. I guess the 3rd point was causing this confusion :)

tsmirnov: One more commit then )

zalmoxisus: Thanks. Except the commit for the 3rd point (or I missed it)? :)

We need it also [here](https://github.com/gaearon/redux-devtools/blob/master/examples/counter/package.json) and [here](https://github.com/gaearon/redux-devtools/blob/master/examples/todomvc/package.json).

tsmirnov: ah... now I got what you mean by 3rd point ) Will fix that as well in a minute

zalmoxisus:
Thanks for the contribution! It's published as `redux-devtools@1.3.0`. Also I published dock and log monitor.

PR #364, opened 2017-05-22
Update Walkthrough.md

Adzz:
Don't we need to compose here? I think we need to use compose but I'm not sure.

If not I can remove it from the imports. I would be curious why not though.


zalmoxisus:
Probably you got your answer long ago already. Composing is for enhancers part only.

PR #372, opened 2017-06-26
Add a label to the 'toggle-all' checkbox in todomvc

danfooo:
Questionable feature for a ToDo list 😉, but since it's implemented it should be probably be rendered.

It would be easier if the label could contain the checkbox, it would work fine without the id, but that's not how the styling is done [in todomvc-app-css](https://github.com/tastejs/todomvc-app-css/blob/master/index.css#L122).


zalmoxisus:
Thanks for contribution!

Issue #381, opened 2017-09-12
React 16


wojtekmaj:
Hey there,
are you planning to officially support React 16?
I have tested this library with 16.0.0-rc.2 and so far, so good.

I'd appreciate feedback from other users how well redux-devtools behaves on their React 16-based environments.

If the feedback will be positive, updated redux-devtools stating React 16 compatibility would be great.

This should be alright IMO:
`"react": "^0.14.9 || ^15.0.0 || ^16.0.0",`

DMShamonov:
Hi @wojtekmaj
With "react": "16.0.0" I have this warnings in console:

> Warning: Expected `onClick` listener to be a function, instead got a value of `boolean` type.
    in label (created by JSONNestedNode)
    in li (created by JSONNestedNode)
    in JSONNestedNode (created by JSONObjectNode)
    in JSONObjectNode (created by JSONNode)
    in JSONNode (created by JSONNestedNode)
    in ul (created by JSONNestedNode)
    in li (created by JSONNestedNode)
    in JSONNestedNode (created by JSONObjectNode)
    in JSONObjectNode (created by JSONNode)
    in JSONNode (created by JSONTree)
    in ul (created by JSONTree)
    in JSONTree (created by LogMonitorEntry)
    in div (created by LogMonitorEntry)
    in div (created by LogMonitorEntry)
    in LogMonitorEntry (created by LogMonitorEntryList)
    in div (created by LogMonitorEntryList)
    in LogMonitorEntryList (created by LogMonitor)
    in div (created by LogMonitor)
    in div (created by LogMonitor)
    in LogMonitor
    in div (created by Dock)
    in div (created by Dock)
    in div (created by Dock)
    in Dock (created by DockMonitor)
    in DockMonitor (created by Connect(DockMonitor))
    in Connect(DockMonitor) (created by DevTools)
    in DevTools (created by AppComponent)
    in div (created by AppComponent)
    in Provider (created by AppComponent)
    in Router (created by BrowserRouter)
    in BrowserRouter (created by AppComponent)
    in AppComponent 

> Warning: Expected `onClick` listener to be a function, instead got a value of `boolean` type.
    in span (created by JSONNestedNode)
    in li (created by JSONNestedNode)
    in JSONNestedNode (created by JSONObjectNode)
    in JSONObjectNode (created by JSONNode)
    in JSONNode (created by JSONNestedNode)
    in ul (created by JSONNestedNode)
    in li (created by JSONNestedNode)
[...]

PR #385, opened 2017-10-12
Added React v16 peer dependency support

jayfunk:
After upgrading to React 16 the redux devtools throws a peerDependency warning.

`npm WARN redux-devtools@3.4.0 requires a peer of react@^0.14.9 || ^15.3.0 but none was installed.`

This fix it to stop this warning and allow downstream modules to build without peerDependency failures.

Linking to #381


mxl:
What does prevent us from merging this PR?

springuper:
We should merge it asap if all tests pass.

ezmiller:
Great that this is done. When will it be installable via npm? Looks like v3.4.0 doesn't have these commits?

mac-s-g:
@ezmiller 3.4.1 is published.  and i can verify that it fixes the issue.

PR #392, opened 2018-01-20
Update .travis.yml

JaKXz:
Just figured it was time to move up :)

Maybe just `"stable"` is enough for this project, to keep build times down?


zalmoxisus:
Thanks for contribution!

Issue #399, opened 2018-04-19
Is Redux 4 officially supported yet?


nickmccurdy:
Redux 4 is now stable. #391 seems to have been fixed and the extension seems to work in Redux's todos example using Redux 4, but the package.json still shows a peer dependency of Redux `^3.5.2`. Is it safe to adjust the peer dependency to allow Redux 4 or are there still changes that should be made to support it? Is there anything we can do to help like testing or updating APIs?

PR #400, opened 2018-04-21
Support Redux 4 in peer dependencies

nickmccurdy:
See #399. The tests appear to work and if I update the other peerDependencies I can get it to install production dependencies without warning. I'm assuming Redux 4 doesn't break anything since the extension still works for me and the tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
   "peerDependencies": {
     "react": "^0.14.9 || ^15.3.0 || ^16.0.0",
     "react-redux": "^4.0.0 || ^5.0.0",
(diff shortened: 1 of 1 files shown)


nickmccurdy:
Note that `redux-devtools-instrument` hasn't had a release since [fixing Redux 4](73ae9765a76adb5efec205412175f66805fb99ca), we might want to wait for that.

brneto:
When are the team planning to release this peerDependency update change?

zalmoxisus:
Thanks for contribution! It's available in `3.4.2`.

PR #402, opened 2018-05-10
Include Typings in package and fix missing instrument options

dtweedle:
So right now typescript users are presented with an error when trying to specify instrument options in the typings.

DefinetlyTyped is also an obsolete monolithic mess and splits away typing concerns from projects making type safety a second class citizen for most libs.

This pull request attempts to address both these issues.

Note that the typings work for redux v4.0.0 and subsequently may not be appropriate to publish as of yet.

I can always downgrade these typings to match the current DT typings and just include the missing options.

Thanks!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,4 @@
+// Redux version 4.0.0
 // Type definitions for redux-devtools 3.4.1
 // TypeScript Version: 2.8.1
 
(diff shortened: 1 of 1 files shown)


zalmoxisus:
Thanks for contribution! It's available in `3.4.2`.

Issue #412, opened 2018-11-17
Monorepo


zalmoxisus:
As discussed before, moving all `redux-devtools` related repositories to `reduxjs` organization doesn't look sustainable (though it's still open for discussion). A better way, especially for new contributors, would be to have everything in a monorepo. We could do that in `redux-devtools-extension` as discussed in https://github.com/zalmoxisus/redux-devtools-extension/issues/412 since there's more activity, but using just `redux-devtools` name sounds better and everything started here. So the plan is to keep only this repository, the rest will be archived and the code moved here. I'll add credits for every readme. Maybe we could even add authors of the monitors in the extension, just need an UI for that (something like a monitors store).

The list of repos to be merged (some will be renamed as suggested bellow):
- [x] `redux-devtools` (from current repo)
- [x] [`redux-devtools-instrument`](https://github.com/zalmoxisus/redux-devtools-instrument)
- [x] [`react-json-tree`](https://github.com/alexkuz/react-json-tree)
- [x] [`redux-devtools-inspector`](https://github.com/alexkuz/redux-devtools-inspector)
- [x]  [`redux-devtools-log-monitor`](https://github.com/gaearon/redux-devtools-log-monitor)
- [x] [`redux-slider-monitor`](https://github.com/calesce/redux-slider-monitor)
- [x] [`redux-devtools-test-generator`](https://github.com/zalmoxisus/redux-devtools-test-generator)
- [x] [`d3tooltip`](https://github.com/romseguy/d3tooltip)
- [x] [`map2tree`](https://github.com/romseguy/map2tree)
- [x] [`devui`](https://github.com/zalmoxisus/devui)
- [x] [`remotedev-utils`](https://github.com/zalmoxisus/remotedev-utils) -> `redux-devtools-core`
- [x] [`remotedev-app`](https://github.com/zalmoxisus/remotedev-app) -> `redux-devtools-core`
- [x] [`remotedev-server`](https://github.com/zalmoxisus/remotedev-server) -> `redux-devtools-cli`
- [x] [`redux-devtools-extension` npm package](https://github.com/zalmoxisus/redux-devtools-extension/tree/master/npm-package)
- [x] [ browser extension](https://github.com/zalmoxisus/redux-devtools-extension)

These repositories could still be maintained outside as either having a maintainer or no changes needed in future (?):
- [x] [`d3-state-visualizer`](https://github.com/reduxjs/d3-state-visualizer/commits/master) (?)
[...]

PR #416, opened 2018-12-07
Monorepo initial configuration

zalmoxisus:
I made a minimal implementation for the monorepo with lerna and using [yarn workspaces](https://yarnpkg.com/en/docs/workspaces), added scripts to run linting and tests inside packages, and for publishing. Probably would be great to have the same linting rules and dev dependences among packages, but let's keep it after moving them all.

The plan is to move packages gradually. For now we have the main part:
- `redux-devtools` (from current repo)
- [`redux-devtools-instrument`](https://github.com/zalmoxisus/redux-devtools-instrument)

EDIT: Moved the list of other packages to #412, to keep the track there.


rufusraghunath:
Hey, as discussed in [zalmoxisus/redux-devtools-extension/issues/468](https://github.com/zalmoxisus/redux-devtools-extension/issues/468), I've had a look at these changes. They look good to me, and the tests are passing.

A few thoughts:

> Not sure if moving all the tests in the root part of monorepo instead of having tests inside for every package is better.

I think your current approach with `run-in-packages.js` is better. Keeping the tests in the individual packages makes it easy to develop the packages and run their tests separately. Just FYI, Jest recursively looks for nested test files out of the box, so if you ever wanted to migrate to Jest, we could remove the Node script :)

> Still unsure about publishing and how to proceed with release tags, redux-devtools is at 3.x while it will be just a package and release tags should be mainly about extension, which is 2.x.

For the release tags, I think the only thing we can do is go to the highest version. So either we go to the 3.x version in `redux-devtools`, or we go straight to 4.0. It's annoying because it breaks semver, but I don't know of any way around this.

For publishing, we should be able to use Lerna's [publish](https://github.com/lerna/lerna/tree/master/commands/publish#readme) flow. That part is usually pretty straightforward. We can use [Verdaccio](https://github.com/verdaccio/verdaccio) as a local NPM registry to safely test publishing. I found this very helpful last time I worked on a monorepo.

> We can improve this later, I'll just move repos gradually. If you'll have time PRs on improving that would be much appreciated.

OK, I can add try moving some more packages when I have time and make a PR to this branch.

zalmoxisus:
@rufusraghunath thanks for helping with it!

> Just FYI, Jest recursively looks for nested test files out of the box, so if you ever wanted to migrate to Jest, we could remove the Node script :)

I think jest will run from the directory of the root package and won't find some dependences which are installed only inside packages (or have different versions there). I guess `lerna test` could help, but that command is not documented, so might be a reason for that.

> For the release tags, I think the only thing we can do is go to the highest version. So either we go to the 3.x version in `redux-devtools`, or we go straight to 4.0. It's annoying because it breaks semver, but I don't know of any way around this.

I think we could go for `3.x` as a canary/nightly version (since we cannot publish alpha versions on Chrome Store and Mozilla Addons) and `4.x` as stable for the extension. But jumping straight from `2.x` to `4.x` would be pretty confusing. So, I guess the best option is to remove tags from this repository at all and start from `3.x`. Those tags referred only for one package, so we might just copy them to `./packages/redux-devtools/CHANGELOG.md`. I think there should be some automated library to copy release tags into a file. I've seen some before but didn't look into.

> For publishing, we should be able to use Lerna's publish flow. That part is usually pretty straightforward. We can use Verdaccio as a local NPM registry to safely test publishing. I found this very helpful last time I worked on a monorepo.

Thanks for the hint. Definitely that would be useful at least for testing first release.

> OK, I can add try moving some more packages when I have time and make a PR to this branch.

I'm still thinking on an option to move all the contribution history here. The problem is that packages commits conflict with the current repository, and after solving conflicts we're losing the contribution. Might try to make a script to override the history so it will look like everything was in `packages/package-name` directory from start.

I think it's good to merge this and have separate PR for every package.

In case others want to help too, the most needed help for now would be with upgrading dependences (as most of packages weren't changed for more than 2 years) and adding tests. Also helping with docs is always appreciated. For now we have docs only for `redux-devtools`, but we need to include information about every package there. The extension and its docs will come at the end.

rufusraghunath:
> I think it's good to merge this and have separate PR for every package.

Won't this block other releases until all the monorepo stuff is done? At least until we have the publishing part sorted?

> I guess that will work only if we have all dependences in the root package and not different versions, otherwise we need to run it from inside the package.

I think this works even if you have separate `devDependencies`. I did this on my last Lerna project... as long as you have a version of Jest installed in the root, it should be able to find all the nested tests and run them.

zalmoxisus:
@rufusraghunath it'll block other releases if we intend to have the same version for every package on publishing, but I changed `version` to `independent` in the commit above (please let me know if you find something wrong there), so we'll have different versions (for now `1.x` for `redux-devtools-instrumentation` and `3.x` for `redux-devtools`). The releases tag version will be for browser extension only. Similarly how it's on [`create-react-app` releases](https://github.com/facebook/create-react-app/releases), the version tag reflects only the main package, while the others specified there have different versions.

zalmoxisus:
Merging this will let us move forward here instead of having changes for current packages outside (like in https://github.com/zalmoxisus/redux-devtools-instrument/pull/24). Some packages like `redux-devtools-app`, `redux-devtools-core` and `redux-devtools-cli` need complete rewrite before merging here. Also there will be new packages, like `redux-devtools-trace-monitor`, which I'll decouple from https://github.com/zalmoxisus/remotedev-app/pull/43.

The plan is to release version `2.17` of the extension, which will implement showing stack traces, since @markerikson did all the work and it's waiting for 3 months already. That will be the last minor there (except some patches for found bugs), all the next features will be implemented here as `3.x`. All issues, except feature requests, were solved in `2.16` during these weeks.

I'll move the list in #412 and we can keep track of what was moved.

rufusraghunath:
> If you'll have some time later, that would be a great contribution, moving tests to jest.

Sure. It probably makes sense to wait until all the packages are moved, though, because it sounds like a lot of them are still being rewritten.

Apart from that, I'm not sure how helpful I can be. You asked for support with testing, but I don't really have much context on this project, so it would be hard for me to write tests. I'm happy to help with the Jest migration later, though.

zalmoxisus:
Thanks, also for the feedback here!

PR #427, opened 2018-12-20
Compatibility with react-redux@6

### LINKED ISSUE #417: Compatibility with react-redux@^6.0.0
Because `redux-devtools@^6.0.0` is no longer supporting passing the `store` as a prop to connected components, using `redux-devtools` is no longer working (when [manually integrating as detailed here](https://github.com/reduxjs/redux-devtools/blob/master/docs/Walkthrough.md#manual-integration)), but instead throwing an error

```
Uncaught Error: Passing redux store in props has been removed and does not do anything. To use a custom Redux store for specific components,  create a custom React context with React.createContext(), and pass the context object to React-Redux's Provider and specific components like:  <Provider context={MyContext}><ConnectedComponent context={MyContext} /></Provider>. You may also pass a {context : MyContext} option to connect
    at invariant (browser.js:38)
    at new Connect (connectAdvanced.js:137)
    at constructClassInstance (react-dom.development.js:12484)
    at updateClassComponent (react-dom.development.js:14255)
    at beginWork (react-dom.development.js:15082)
    at performUnitOfWork (react-dom.development.js:17820)
    at workLoop (react-dom.development.js:17860)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
```

zalmoxisus:
Fixes #417.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,5 @@ cache:
     - "node_modules"
 script:
   - npm run lint
+  - npm run build:all
   - npm test
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "scripts": {
     "lerna": "lerna",
     "prepare": "lerna run prepare --since master --stream --sort -- --scripts-prepend-node-path",
+    "build:all": "lerna run build",
     "publish": "lerna publish",
     "canary": "lerna publish --canary preminor --npm-tag alpha",
     "next": "lerna publish --bump prerelease --npm-tag next",
--- a/packages/redux-devtools/examples/counter/package.json
+++ b/packages/redux-devtools/examples/counter/package.json
@@ -17,12 +17,12 @@
   "homepage": "https://github.com/reduxjs/redux-devtools#readme",
   "dependencies": {
     "prop-types": "^15.5.7",
-    "react": "^15.3.0",
-    "react-dom": "^15.0.1",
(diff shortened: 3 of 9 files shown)


Ephem:
Looks great!

Contrary to my comment in #417 I like the use of Consumer here (since `contextType` for new context is only supported from 16.6). When looking at your solution it feels obvious in hindsight that using the `<Provider>`-component from `react-redux` directly is the best solution (since this is dealing with a separate store).

If someone is using a custom context for `react-redux`, they can pass the store in directly via the prop, so this PR also supports that usecase.

Nice work! :clap:

zalmoxisus:
@Ephem thanks for looking into it!

> If both stores updates at once, I guess each update might trigger a separate render? 
Except when updating a monitor state, for every dispatch, everytime both stores are updated simultaneously in [instrumentation](https://github.com/reduxjs/redux-devtools/blob/master/packages/redux-devtools-instrument/src/instrument.js). That's a valid concern. I removed `shouldComponentUpdate` [here](https://github.com/gaearon/redux-devtools-log-monitor/blob/master/src/LogMonitor.js#L70) and still there's only one render for every update.

zalmoxisus:
Published as `redux-devtools@3.5.0`.

PR #434, opened 2018-12-22
Add redux-slider-monitor

jhen0409:
Related to #412.

@zalmoxisus I also added you as a collaborator of `redux-slider-monitor` NPM package.


zalmoxisus:
Yay, thanks! I just wanted to ask you about that 😀

zalmoxisus:
@jhen0409 added you to the collaborators here. I cannot add members to the org, but that shouldn't be necessary anyway if we have everything in this monorepo.

PR #440, opened 2018-12-26
add website folder created with docusaurus 

solo474:
Adding a website folder created with docusaurus
Can be deployed to Netify


zalmoxisus:
We used gitbook for the extension, but docusaurus would be better to have the same format as other reduxjs projects. Thanks for helping with it!

PR #445, opened 2019-01-03
Move remotedev-app and remotedev-server

zalmoxisus:
Implement https://github.com/zalmoxisus/remotedev-app/pull/30 into `redux-devtools-core` and create `redux-devtools-cli` and  as per #412.


zalmoxisus:
Looks like `redux-devtools-app` is taken. We'll better include the ui part in `redux-devtools-core` as well, anyway it would depend on it in future. The electron app will be included in `redux-devtools-cli`.

PR #449, opened 2019-01-10
Fix and update packages

nndio:


zalmoxisus:
Thanks for all the work here!

PR #456, opened 2019-01-26
fix(filter-keys): Filter state by keys,v1

PaulRosset:
Hi :)

As an answer to the previous opened [!617](https://github.com/zalmoxisus/redux-devtools-extension/issues/617) issue, here is my implementation of the requested feature.

- I'm using the library [flat](https://www.npmjs.com/package/flat), to flat the whole state object and get the keys path of each value in the state.
- `flat` function return an object, but I need an array that would give me possibility to after apply some high order function on it, so, I made a little util function `objectMap` to transform my object into an array of the form: `[{value, key}]`
- Then, I can apply a `filter` function, to select only the matched string.
- Finally, I'm doing a `reduce` function to display the result in a correct manner in the view.

This is my first iteration, tell me what do you think :)

### DIFF
--- a/packages/redux-devtools-inspector/package.json
+++ b/packages/redux-devtools-inspector/package.json
@@ -73,6 +73,7 @@
   "dependencies": {
     "babel-runtime": "^6.3.19",
     "dateformat": "^1.0.12",
+    "flat": "^4.1.0",
     "hex-rgba": "^1.0.0",
     "immutable": "^3.7.6",
     "javascript-stringify": "^1.1.0",
--- a/packages/redux-devtools-inspector/src/tabs/StateTab.jsx
+++ b/packages/redux-devtools-inspector/src/tabs/StateTab.jsx
@@ -1,18 +1,93 @@
-import React from 'react';
+import React, { Component } from 'react';
 import JSONTree from 'react-json-tree';
 import getItemString from './getItemString';
 import getJsonTreeTheme from './getJsonTreeTheme';
+import objectMap from '../utils/objectMap';
+import flatten from 'flat';
 
-const StateTab = ({
-  nextState, styling, base16Theme, invertTheme, labelRenderer, dataTypeKey, isWideLayout
-}) =>
-  (<JSONTree
-    labelRenderer={labelRenderer}
-    theme={getJsonTreeTheme(base16Theme)}
-    data={nextState}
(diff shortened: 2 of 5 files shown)


zalmoxisus:
Hey! That looks great. Thanks for working on it!

Would it be possible to reuse it for Action tab as well? And also it could be useful for Diff tab in some cases. Probably you could just use `keyPath` prop of [`react-json-tree`](https://github.com/reduxjs/redux-devtools/tree/master/packages/react-json-tree). There's no description for it unfortunately, but from looking to the source I see we could just split that found flatted path as an array and pass to it.

Another solution would be to use `labelRenderer`. Also it would be useful to filter values, for which there's `valueRenderer`. [`redux-devtools-filterable-log-monitor`](https://github.com/bvaughn/redux-devtools-filterable-log-monitor/) is a good example for the latter, but not sure how much it'll complicate things.

PaulRosset:
Hey, I'm wondering if it will be useful for the Action and Diff tab, because flow of data that go through these tabs in redux are small most of the time.  
What do you think ?

zalmoxisus:
For Diff yes, but for the Action there're often large payloads, where you're interested only by a small part. It's also useful to see the same key for action and state object. Then if we'd have for those 2 tabs, shouldn't be difficult to add for Diff as well, where one could be interested only in a subpart of the diffed object.

I might be mistaken about `react-json-tree` capabilities, since I didn't contribute much to that package, but if it has such functionality to filter keys, we better use it, instead of adding that again.

xmile1:
Hi, thanks for the work, please is there any hope to get this feature merged or is this already implemented somewhere else ?
cc: @PaulRosset @zalmoxisus

PaulRosset:
@xmile1  Would love to see this PR merged.

Its been quite a long time now that the PR is opened, I will dive again into asap, to see if we can improve it to be merged.

alexmclaurian:
@PaulRosset this is a great idea and would be helpful, have there been any updates on this?

Methuselah96:
Feel free to resolve the merge conflicts in this PR or create a new PR and I'll review it and merge it if it's good.

Methuselah96:
Closing since this has been stale for a few years and these files have been converted to TypeScript since then. Feel free to re-open or open a new PR if you would like to pursue this further.

PR #458, opened 2019-02-03
[redux-devtools-trace-monitor] StackTraceTab: Allow custom `openFile` function

jhen0409:
I'm trying to integrate the trace feature in [react-native-debugger](https://github.com/jhen0409/react-native-debugger/pull/301), and I think it also would be helpful in standalone app of [remotedev-app](https://github.com/zalmoxisus/remotedev-app/blob/master/src/app/containers/monitors/InspectorWrapper/index.js).

For follow-up work, we need make some changes in `remotedev-app`.

### DIFF
--- a/packages/redux-devtools-trace-monitor/src/StackTraceTab.js
+++ b/packages/redux-devtools-trace-monitor/src/StackTraceTab.js
@@ -7,6 +7,9 @@ import openFile from './openFile';
 const rootStyle = {padding: '5px 10px'};
 
 export default class StackTraceTab extends Component {
+  static defaultProps = {
+    openFile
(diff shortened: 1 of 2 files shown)


zalmoxisus:
That is great to have. Thanks! Published `redux-devtools-trace-monitor@0.1.1`.

zalmoxisus:
BTW, `remotedev-app` is being moved to `redux-devtools-core`, and electron app to `redux-devtools-cli`, which will be opened with [`--open` argument](https://github.com/reduxjs/redux-devtools/tree/master/packages/redux-devtools-cli#open-redux-devtools).

PR #472, opened 2019-04-30
Allow custom keys for slider monitor

### LINKED ISSUE #471: [redux-slider-monitor] Allow overriding keyboard hotkeys via props
redux-slider-monitor should allow configuring hotkeys like redux-devtools-dock-monitor using parse-key.

Currently the key codes are hardcoded: https://github.com/reduxjs/redux-devtools/blob/master/packages/redux-slider-monitor/src/SliderMonitor.js#L83-L104

okorz001:
This is mostly lifted from [redux-devtools-dock-monitor](https://github.com/gaearon/redux-devtools-dock-monitor/blob/848eb32dd7b76db0b8d191bb447d39bc6376af59/src/DockMonitor.js#L67-L113), but there was some extra complexity around the fact that the underlying parse-key library doesn't support ctrl-[ and ctrl-]. Ideally, I would have defaulted the new props to the default bindings, but I had to settle for falling back to original logic when they are unset.

I couldn't get the example todomvc project running, but I was able to verify these changes in my own project, both with and without the new props.

Closes #471.

### DIFF
--- a/packages/redux-slider-monitor/README.md
+++ b/packages/redux-slider-monitor/README.md
@@ -43,6 +43,8 @@ Pass the `keyboardEnabled` prop to use these shortcuts
 
 `ctrl+]`: step forward
 
+You can override these shortcuts with the `playKey`, `stepBackwardKey`, and `stepForwardKey` respectively. These props are parsed by [parse-key](https://github.com/thlorenz/parse-key), just like in `DockMonitor`.
+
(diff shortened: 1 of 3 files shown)


okorz001:
The build failure doesn't have anything to do with my changes. There's some node-gyp failures, probably related to node 12 being released.

PR #476, opened 2019-06-09
[redux-devtools-inspector]: filter - use '!' to negate query

adekbadek:
**The problem:**
When using the devtools, I missed the functionality to filter by _not matching_ query, e.g. to just show actions which are not of certain type, to reduce clutter when debugging.

**The solution:**
This initial implementation proposes `!` prefix as a way to negate a query. So `!ADD` would filter _out_ any actions that contain the word `ADD` in their type.

**Problem with the solution**
The only drawbacks I see now are:
1. discoverability - this would of course be documented, but maybe a button would be a better solution from a UX perspective? 
2. actions with `!` as first character of `type`. This is probably a super edge-case, but maybe that would be a problem for someone.

Let me know what you think, I'd be happy to work more on that feature (docs, tests, etc)!

### DIFF
--- a/packages/redux-devtools-inspector/src/ActionList.jsx
+++ b/packages/redux-devtools-inspector/src/ActionList.jsx
@@ -94,13 +94,19 @@ export default class ActionList extends Component {
       onJumpToState
     } = this.props;
     const lowerSearchValue = searchValue && searchValue.toLowerCase();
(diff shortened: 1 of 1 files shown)


adekbadek:
Hi @zalmoxisus , did you get a chance to look at this?

PR #482, opened 2019-08-26
Update react-redux peer dependency

leMaik:
We have a project with react-redux@7 and as far as I can see, redux-devtools work just fine. This PR fixes the peer dependency warning npm gives when installing the package, reducing CLI noise.


tommarien:
Can someone please check the build server, it looks failing node-gyp build for sqllite3 (on node V12 ;))

AndreaCitrolo:
Hi,
is it safe to use redux-devtools with react-redux@7+ or is the build failing due to incompatibilities?

leMaik:
@AndreaCitrolo It works just fine :+1:

timdorr:
Thanks!

PR #483, opened 2019-08-29
[react-json-tree] Mark deprecated React lifecycle methods as unsafe

Dehorser:
This will allow this package to be updated to React 17. For more information, see: https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

### Fixes: https://github.com/reduxjs/redux-devtools/issues/480


timdorr:
Thanks!

PR #485, opened 2019-09-24
Fix broken link

mottox2:
Fix broken link in README.md


timdorr:
Thanks, but this should be a relative link instead of absolute.

PR #564, opened 2020-08-07
[redux-devtools-cli] Forward `port` flag to Electron

geekjuice:
### Description

When running `redux-devtools` with the `open` flag, the `port` option is ignored in `app/electron.js` as the `spawn` command does not forward the flag from the parent command.

The changes here ensures that the `port` option is forwarded from the `open.js` script so that it can be properly used in the `electron.js` app.


Methuselah96:
Thanks!

PR #627, opened 2020-09-03
Don't mutate source object during stringification

### LINKED ISSUE #625: Data is mutated during stringification
See https://github.com/reduxjs/redux-devtools/blob/c7b0c7aa6e09f46a36b382ae3ec8e38bd48aeb28/packages/redux-devtools-core/src/app/utils/stringifyJSON.js#L7

Sadly, it would seem the original author has passed away or has otherwise checked out of the project, so I won't labour the point, but this is pretty poor form. This mutation causes the `__serializedType__` annotation to be unusable for correctly and consistently reviving complex types during time travel.

wms:
Fixes #625 

The underlying issue is reproduced via a test case in b3083e5. Then, in 
c04f28e, we use an object spread to produce a shallow clone of `value`, so that we `delete` a key from it without affecting the source.

I opted for using the object spread operator since it leverages language builtins, and am assuming that either:
- We're only interested in targetting environments where object spread is supported
- We target environments where object spread _is not_ supported, but Babel and friends are configured to downlevel it correctly.


Methuselah96:
Thanks!

Issue #635, opened 2020-09-16
React-json-tree: Update UNSAFE_componentWillReceiveProps lifecycle hook.


BrunoFenzl:
I could not find any related issues, so please forgive me if this was already addressed somewhere.
React throws following warning when using `react-json-tree`:
 
```
Warning: Using UNSAFE_componentWillReceiveProps in strict mode is not recommended and may indicate bugs in your code. See https://fb.me/react-unsafe-component-lifecycles for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state

Please update the following components: JSONNestedNode, JSONTree
```

I believe I could fix this and update those components. Is this in your roadmap?

Methuselah96:
What version of `react-json-tree` are you using? It should be fixed in [`react-json-tree@0.12.0`](https://github.com/reduxjs/redux-devtools/releases/tag/react-json-tree%400.12.0).

BrunoFenzl:
I'm using version 0.13.0. The PR #483 just updates the deprecation messages using `UNSAFE_componentWillReceiveProps` instead of implementing `getDerivedStateFromProps`. I propose updating those comps with the new static method.

Methuselah96:
Oh, I see, sorry, I read it the issue too fast. Yeah, it's worth fixing, so PRs are welcome. Eventually I will get to it myself, but it's not the highest priority right now.

PR #644, opened 2020-09-25
Update unsafe_componentWillReceiveProps react-json-tree

BrunoFenzl:
This PR attempts to solve #635. 
`unsafe_componentWillReceiveProps` is currently used in two components: `JSONTree` and `JSONNestedNode`.

In `JSONTree` I extracted the properties `theme` and `invertTheme` from `Props` to a separate interface and extended the `State` and `Props` interfaces with it. The reason being, that the new static method `getDerivedStateFromProps` does not have access to the current props, so properties that need to be compared should be mirrored in the local state object.

In `JSONNestedNode` there is no need to check for derived state at all. The property `expanded` can be copied to state and re-rendering is checked anyway in the `shouldComponentUpdate` lifecycle hook.

I'm looking forward to feedback and will be glad to refactor if needed.


Methuselah96:
@BrunoFenzl Thanks for making a PR!

1. `JSONNestedNode`: `getStateFromProps` is now only ever called in the `constructor` when before it was called on every render. I think you might have to call `getDerivedStateFromProps` in order to match the previous behavior unless I'm missing something. I would like to refactor it to remove `expanded` from `state` altogether, but I don't think that's possible without calling `shouldExpandNode` twice since `shouldComponentUpdate` depends on it.
2. `JSONTree`: The React team recommends not copying props to state just to memoize a value. I would prefer using `memoize-one` as suggested in [this blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization).

BrunoFenzl:
@Methuselah96 Thanks for your feedback!

1. You're right. I had `getDerivedStateFromProps` initially in place but thought it would be redundant because of the check in 
 `shouldComponentUpdate`. I see it now I was clearly wrong.
2. I am aware of the article but didn't want to add another dependency. If it is ok for you, I'll refactor to use memoize-one.

Methuselah96:
@BrunoFenzl Yeah, I'm fine with adding `memoize-one` as a dependency. Thanks!

dstiers11 (approved):

yesh0907:
Hi, can this PR be merged? Thanks.

Methuselah96:
The feedback I gave still needs to be addressed. Feel free to create your own PR since it seems like this PR might be abandoned.

defunctzombie:
@Methuselah96 Would you be open to a PR which removes these unsafe calls by converting the JSONTree and JSONNestedNode to function components and using hooks?

Methuselah96:
@defunctzombie Yeah, go for it!

Methuselah96:
Resolved by https://github.com/reduxjs/redux-devtools/pull/1288.

PR #660, opened 2020-10-30
use latest symbol-observable in redux-devtools-instrument

EtDu:
Latest version is compatible with environments where Symbol may be frozen.


Methuselah96:
Thanks!

PR #685, opened 2020-12-03
remove is required check

chriswoodley:
See this issue https://github.com/reduxjs/redux-devtools/issues/684


Methuselah96:
Thanks!

PR #694, opened 2020-12-22
feat(react-json-tree): Add keyPath to getItemString

joshhunt:
Adds `keyPath` to `getItemString` so I can conditionally render different labels depending on the path.

Same as https://github.com/reduxjs/redux-devtools/pull/470, but updated to Typescript. Credit goes to @slavasitnikov, I just ported their changes to the new Typescript.


Methuselah96 (approved):
Thanks!

joshhunt:
Thank you @Methuselah96! Now I can provide annotations for object keys!

image

😄

PR #709, opened 2021-04-25
fix(devui): Fix buggy UI behaviors with redux-devtools-app

hanayashiki:
1. The InspectionWrapper cannot jump within Action, State, and Diff. This behavior is due to `Tabs` in devui only reacts to changes of prop `selected`. When Action, State, and Diff share the same subtab, the changes cannot be reflected inside `Tabs`.  This problem is fixed by always recalculating `updateTabs` in `Tabs`.
image

2. When the action list grows, it makes the whole container grow outside the window.
Reported in: https://github.com/reduxjs/redux-devtools/issues/699. Fixed by adding `flex-shrink: 1` to the container.

This app is really feature rich but sadly these problems seem to remain there for a very long time. Looking forward to your response.


hanayashiki:
By the way, I made an online version for it

https://redux-devtools-app.vercel.app/

packages/devui/src/Tabs/Tabs.tsx
```diff
@@ -24,10 +24,8 @@ export default class Tabs<P> extends Component<TabsProps<P>> {
   SelectedComponent?: React.ComponentType<P>;
   selector?: () => P;
 
-  UNSAFE_componentWillReceiveProps(nextProps: TabsProps<P>) {
-    if (nextProps.selected !== this.props.selected) {
-      this.updateTabs(nextProps);
-    }
+  UNSAFE_componentWillReceiveProps(nextProps: TabsProps<P>): void {
+    this.updateTabs(nextProps);
   }
```
Methuselah96, line 29: If we're going to unconditionally update the tabs can we go ahead and just turn `updateTabs()` into `renderTabs()` and call it from the `render()` method?

packages/devui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "devui",
-  "version": "1.0.0-8",
+  "version": "1.0.0-9",
```
Methuselah96, line 3: Can you revert the version bump? I'll bump it when I got to deploy it.

packages/redux-devtools-app/package.json
```diff
@@ -45,7 +45,7 @@
     "@redux-devtools/log-monitor": "^2.3.0",
     "@redux-devtools/slider-monitor": "^2.0.0-7",
     "d3-state-visualizer": "^1.4.0",
-    "devui": "^1.0.0-8",
+    "devui": "^1.0.0-9",
```
Methuselah96, line 48: Same here.

Methuselah96 (changes requested):

packages/devui/src/Tabs/Tabs.tsx
hanayashiki: I think so

packages/redux-devtools-app/package.json
hanayashiki: ok

packages/devui/package.json
hanayashiki: ok

packages/devui/src/Tabs/Tabs.tsx
```diff
@@ -67,6 +61,8 @@ export default class Tabs<P> extends Component<TabsProps<P>> {
...
 
   render() {
+    this.updateTabs(nextProps);
```
Methuselah96, line 64: `nextProps` isn't defined here which is why the build is failing. Can you instead turn this into a `renderTabs` and remove the `tabsHeader` class variable altogether? Thanks!

Methuselah96 (changes requested):

packages/devui/src/Container/styles/index.ts
```diff
@@ -6,6 +6,8 @@ export const MainContainerWrapper = styled.div`
   height: 100%;
   width: 100%;
   flex-flow: column nowrap;
+  flex-shrink: 1;
+  min-height: 0;
```
Methuselah96, line 10: Can we use `overflow: auto` instead of `min-height: 0` here and below? Also if I understand the situation correctly, we don't need `flex-shrink: 1` since that's the default.

Methuselah96 (changes requested):

packages/devui/src/Tabs/Tabs.tsx
Methuselah96: Or you can just pass in `this.props` for now if that's easier.

Methuselah96:
I'm just going to make a follow-up PR to clean some of this up. Thanks for the contribution!

PR #750, opened 2021-06-20
[feat][monitor] Add rkt-query-inspector-monitor - feat/rtk query monitor

FaberVitale:
#  PR: feat/rtk query monitor

## Description

This PR adds a new monitor that displays [RTK query](https://redux-toolkit.js.org/rtk-query/overview) queries and mutations for [Redux DevTools](https://github.com/gaearon/redux-devtools).

The idea for this project has started as a discussion at `#redux` channel and has evolved in a discussion in a [redux-toolkit issue](https://github.com/reduxjs/redux-toolkit/issues/1097).

More informations are available at [rtk-query-inspector-monitor readme](https://github.com/FaberVitale/redux-devtools/blob/feat/rtk-query-monitor/packages/redux-devtools-rtk-query-inspector-monitor/README.md)

## Demo

* demo: https://rtk-query-monitor-demo.netlify.app/

### Dock monitor

RTK Query inspector monitor demo

### App - default theme inverted
Screenshot 2021-06-20 at 11 23 26

### App - default theme
Screenshot 2021-06-20 at 11 23 07
---

## Notable repo changes

* added `@redux-devtools/rtk-query-inspector-monitor` to monorepo

* added `packages/redux-devtools/examples/rtk-query-polling`

* integrated `@redux-devtools/rtk-query-inspector-monitor` to `@redux-devtools/app`

* added `@types/react` as monorepo devDep

### `rtk-query-inspector-monitor`

`rtk-query-inspector-monitor` has almost reached all the features planned, [see TODO section for more details](https://github.com/FaberVitale/redux-devtools/blob/feat/rtk-query-monitor/packages/redux-devtools-rtk-query-inspector-monitor/README.md#todo), but feedback and new ideas are welcome.

### Integrations

* Verify extension integration, It wasn't clear how to start a dev server and what 'port is using.

* Verify monitor on browsers that runs on windows: ugly scrollbars? other issues?

* Fix linting errors that have emerged in `redux-devtools-serialize`

* Other?

PR #754, opened 2021-07-25
[feat][inspector-monitr] add noop and rtk action filters  (issue #753 #717) #717

### LINKED ISSUE #717: Suggestion: add filter to hide all "no-op" actions
It's possible for actions to result in no state updates at all. Some actions are dispatched just as "signals" to middleware (common with sagas/observables). Others just get ignored by reducers.

It would be potentially useful to have a filtering option that hides actions that resulted in no state changes.

(suggested over in Reactiflux)

### LINKED ISSUE #753: [feat][inspector-monitor] Add option to toggle visibility of RTKQ actions
[...]
@FaberVitale Can you add a checkbox to the default action monitor to show/hide RTKQ-related actions? I think one of the biggest critiques towards RTKQ in terms of debugging was that it dispatches a lot of actions, and makes non-RTKQ-code more difficult to debug. So having a switch for just hiding those automated actions would be super-cool.
[...]

_Originally posted by @phryneas in https://github.com/reduxjs/redux-devtools/issues/750#issuecomment-882030985_

FaberVitale:
## Description

This PR adds 3 toggle buttons to filter the actions displayed in the `ActionList` section.
A demo of the standalone monitor is available [here](https://redux-devtools-filter-buttons-demo.netlify.app/?dark).

### Noop filter

when active hides actions that haven't caused state updates.
Closes #717 .

### RTKQ filter

when active filters out RTKQ actions.
Closes #753 .

### Invert Search

when active reverse returns all the actions that **do not** match the current search.
It behaves like the `-v --invert-match` flag of grep.
Related PR #476 .

### Other changes

Moves `Commit` and `Sweep` buttons below the input element when `hideMainButtons` is `false`.

---

### Standalone Demo

- https://redux-devtools-filter-buttons-demo.netlify.app/?dark

### Extension pics

before-rtkq-f
after-rtq-f

## Acknowlegments

Thanks to @phryneas  and @markerikson for the original ideas #753 #717.

---

## Related Issues and PR

Closes #753 #717
Related PR #476 #750


markerikson:
I like the idea in principle, but I think we might want to take a different approach here.

Having hardcoded buttons for "Noop", "RTKQ", and "!" seems a little too overly-specific to me.

Instead, I'm going to suggest that we swipe an idea from the React DevTools. I think it might be worth adding a more dynamic set of filtering options that can be toggled on and off, with a sort of "query builder"-style UI:

image

Can we make something like that for filtering out Redux actions?

FaberVitale:
Hi @markerikson,

Thanks for the feedback.

I'm not opposed to the idea of an advanced search form in another view
but I'm not sure If strictly following the react devtools UX is a good idea.

### Toggle buttons

> Having hardcoded buttons for "Noop", "RTKQ", and "!" seems a little too overly-specific to me.

#### RTKQ filter

RTKQ filter is implemented using a regex generated at runtime that is roughly [`/^(api1|api2|...)/(executeQuery|executeMutation|...)/`](https://github.com/reduxjs/redux-devtools/pull/754/files#diff-8867729139e519b94440f0926c8ff02144d68d605c61026ce006ae21b2a6f94eR75)

I do not think that it would be wise to replace that with a generic `regex filter`:
it's rather tedious to write that regex by hand.
Moreover what I've implemented is basically what was asked [here](https://github.com/reduxjs/redux-devtools/pull/750#issuecomment-882030985).
I might convert that to a checkbox/toggle with a clear label but I do not believe that's better to write everytime a regex rather than clicking a button once.

#### Noop Action

I agree that I unfortunately that button label it is not that clear enough  but I hoped that clear `[title]` and `[aria-label]` would have sufficed.
In a wider form it could be replaced by a checkbox with a proper label.

---

### Advanced Search Form

> Can we make something like that for filtering out Redux actions?

I can create an advanced search view that is rendered over the `ActionList`.
I'm not really sure that kind of features we could add tho.

I believe that this advanced search form should keep `rtkq` and `noop` actions,
converted to checkboxes/switches with clearer labels.

####  Form widgets

1. search/regex search
2. an hasIn property path
3. `is FSA` checkbox (maybe)
4. Invert search.

---

Have you got a specific widget in mind?
Feedback?

markerikson:
I'm fine with having a generated set of RTKQ regexes, I'm just proposing that it be one of the selectable things in a filter menu that can then be toggled on and off.

Like, in that React DevTools filter example, have "RTK Query" be one of the selectable options in the dropdown on the left, with no further options on the right.

I still like the React DevTools filter UI approach: the ability to add multiple filters that get combined together, and each filter can be defined using several pre-existing options.

PR #755, opened 2021-08-11
Update README.md links from zalmoxisus to reduxjs

JimiPedros:
- as a new contributor it is confusing being redirected to a separate repository when navigating the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,24 +2,24 @@
 
 # Redux DevTools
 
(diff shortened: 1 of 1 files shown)


Methuselah96 (changes requested):
Thanks! Unfortunately a lot of these links are still incorrect. For example, the correct base URL for the extension is now https://github.com/reduxjs/redux-devtools/blob/master/extension/docs/Integrations.md and `remotedev-app` should be https://github.com/reduxjs/redux-devtools/tree/master/packages/redux-devtools-app.

PR #861, opened 2021-09-14
Feat/use dark mode preference to set initial theme.light

FaberVitale:
### Description

Displays `@redux-devtools/app` and `redux-devtools-extension` in dark mode if  the following conditions are all true:

1. user prefers dark color scheme according to https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme
2. user has not edited applied changes to the default theme using `settings > theme` form.

### DIFF
--- a/packages/redux-devtools-app/jest.config.js
+++ b/packages/redux-devtools-app/jest.config.js
@@ -3,6 +3,6 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/test/setup.ts'],
   testEnvironment: 'jsdom',
   moduleNameMapper: {
-    '\\.css$': '<rootDir>/test/__mocks__/styleMock.ts',
+    '\\.css$': '<rootDir>/test/__mocks__/styleMock.js',
   },
 };
--- a/packages/redux-devtools-app/src/reducers/theme.ts
+++ b/packages/redux-devtools-app/src/reducers/theme.ts
@@ -1,15 +1,25 @@
-import { Scheme, Theme } from '@redux-devtools/ui';
+import { Theme, Scheme } from '@redux-devtools/ui';
 import {
   CHANGE_THEME,
   APPLY_MEDIA_FEATURES_PREFERENCES,
 } from '../constants/actionTypes';
 import { StoreAction } from '../actions';
 
+export const defaultThemeColorPreference = 'default';
+
+export const themeColorPreferences = [
+  defaultThemeColorPreference,
+  'light',
+  'dark',
+] as const;
+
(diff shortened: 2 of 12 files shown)


packages/redux-devtools-app/src/reducers/theme.ts
```diff
@@ -21,7 +25,20 @@ export default function theme(
...
       scheme: action.scheme,
       light: !action.dark,
+      latestChangeBy: CHANGE_THEME,
     };
   }
+
+  if (
+    action.type === APPLY_MEDIA_FEATURES_PREFERENCES &&
+    state.latestChangeBy !== CHANGE_THEME
+  ) {
```
Methuselah96, line 35: Instead of checking on startup can we make it three options: "use system default", "light", or "dark"?

Methuselah96 (changes requested):
Thanks for working on this! Sorry for taking so long to look at it.

packages/redux-devtools-app/src/reducers/theme.ts
FaberVitale: sure but we have got to store this preference in a different property name or it would not work properly due to the way the store data is hydrated.

Nice idea

FaberVitale:
I've implemented the changes requested in https://github.com/reduxjs/redux-devtools/pull/861#discussion_r739652680

Methuselah96 (approved):
Thanks for all your help! 🚀 Sorry for the delayed reviews.

packages/redux-devtools-app/jest.config.js
```diff
@@ -3,6 +3,6 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/test/setup.ts'],
   testEnvironment: 'jsdom',
   moduleNameMapper: {
-    '\\.css$': '<rootDir>/test/__mocks__/styleMock.ts',
+    '\\.css$': '<rootDir>/test/__mocks__/styleMock.js',
```
Methuselah96, line 6: Was it cause problems for this to be `ts`?

FaberVitale: I do not recall exactly, luckily I wrote a detailed commited message about it.

```ts
chore(@redux-devtools/app): fix lint error

Error message:

0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: test/__mocks__/styleMock.ts.
The file must be included in at least one of the projects provided
```

Methuselah96: Nice commit message! I wonder if those was fixed by https://github.com/reduxjs/redux-devtools/pull/867.

PR #862, opened 2021-09-15
chore(*): remove unnecessary immutable flag

Methuselah96:


nickmccurdy:
Out of curiosity why is this unnecessary? I don't see `--frozen-lockfile` in yarn 3.

Methuselah96:
One of the changes in Yarn 3 was "Make Yarn imply --immutable when used on CI" (https://github.com/yarnpkg/berry/issues/1406) so it should be used without manually specifying it now.

nickmccurdy:
Great, I hated having to remember that in v1

PR #864, opened 2021-09-17
chore(*): remove Travis

nickmccurdy:
The monorepo is already using GitHub Actions instead of Travis, but there are still some old configs and broken links


nickmccurdy:
@phryneas

packages/react-base16-styling/README.md
```diff
@@ -1,4 +1,4 @@
-# react-base16-styling [![Build Status](https://img.shields.io/travis/alexkuz/react-base16-styling/master.svg)](https://travis-ci.org/alexkuz/react-base16-styling) [![Latest Stable Version](https://img.shields.io/npm/v/react-base16-styling.svg)](https://www.npmjs.com/package/react-base16-styling)
+# react-base16-styling ![GitHub Workflow Status](https://img.shields.io/github/workflow/status/reduxjs/redux-devtools/CI) [![Latest Stable Version](https://img.shields.io/npm/v/react-base16-styling.svg)](https://www.npmjs.com/package/react-base16-styling)
```
Methuselah96, line 1: For consistency, let's just remove those badges since we're not including them in the other packages.

packages/redux-devtools/README.md
```diff
@@ -14,7 +14,7 @@ A live-editing time travel environment for [Redux](https://github.com/reactjs/re
 - [Custom Monitors](#custom-monitors)
 - [License](#license)
 
-[![build status](https://img.shields.io/travis/reduxjs/redux-devtools/master.svg?style=flat-square)](https://travis-ci.org/reduxjs/redux-devtools)
+![GitHub Workflow Status](https://img.shields.io/github/workflow/status/reduxjs/redux-devtools/CI)
```
Methuselah96, line 17: We can probably drop this one too.

Methuselah96 (changes requested):
Thanks for cleaning this up!

Methuselah96 (approved):
Thanks!

nickmccurdy:
For the future you may want to note that the monorepo doesn't seem to be publishing the `react-dock` demo, though the package hasn't changed much so the old demo link still works.

Methuselah96:
Looks like we'd have to update where the demo is hosted from. My top priority has been getting 3.0 of the extension released, but cleanup of all the READMEs and documentation is next on my list after that.

nickmccurdy:
I searched for all references to the old repositories when making this PR, so the demo should be the only old repository link that needs to be cleaned up.

PR #865, opened 2021-09-17
chore(*): build on Linux for performance

nickmccurdy:
Should work since Travis was using Ubuntu previously


Methuselah96:
I'm curious to see if this works. IIRC the E2E extension tests were not able to run successfully when I which is why [I ended up switching to Windows](https://github.com/reduxjs/redux-devtools/pull/749). I'm sure there's a way to do it (probably involving setting up xvfb) but it was easier just to switch to Windows at the time.

nickmccurdy:
I don't think I can everything locally, so I'll wait for the CI status first. If that ends up being the case, it might be better to have the E2E tests in a separate job.

Methuselah96:
Yeah, the old repo [was using AppVeyor](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/appveyor.yml), presumably for the same reason.

Methuselah96:
Yeah, I'm okay with a separate job, but there would definitely be duplicate work since the E2E tests rely on building almost all of the packages which is the slowest part of the build, but I suppose it would overall be faster if the Windows didn't have to run the other stages of the build. My preference would be to get the E2E tests working on Linux.

nickmccurdy:
That sounds good, but the build is still failing on the lint step. Any idea why?

Methuselah96:
https://github.com/reduxjs/redux-devtools/pull/867 should fix the lint.

Methuselah96:
Hope you don't mind I added to the PR. Thanks for you work on this!

Issue #995, opened 2022-01-08
Feedback on new v3.0 UI


markerikson:
Per discussion in Reactiflux, congrats on getting this out the door!

My main nitpick, as I noted earlier, is that all of the "data" areas of the UI (actions list, action/state tab contents, RTKQ tab) are using a variable-width font, regardless of whether the main theme setting is "Default" or "Material".  In this case it looks like the font is `Source Sans Pro`:

![image](https://user-images.githubusercontent.com/1128784/148651369-fa2ee09c-ae58-4392-b931-b1602a7e1d39.png)

I _personally_ would really prefer to still have all the actual data in a fixed-width / monospace font.

It would also be helpful to have a font size setting for the content areas, and _maybe_ bump up the default font size for the content just a bit.

Interestingly, it looks like the "Log Monitor" _is_ using a monospace font, "Source Code Pro":

![image](https://user-images.githubusercontent.com/1128784/148651385-24c20c74-02dc-4c31-9797-12b10bd74db8.png)

Also: the "tabs"-ish buttons on the main panel are capitalized ("Actions", "State"), but the buttons on the RTKQ panel are all lower-case ("query", "api").

FaberVitale:
> Also: the "tabs"-ish buttons on the main panel are capitalized ("Actions", "State"), but the buttons on the RTKQ panel are all lower-case ("query", "api").

My bad :)

I'll make a PR.

jjspace:
> My main nitpick, as I noted earlier, is that all of the "data" areas of the UI (actions list, action/state tab contents, RTKQ tab) are using a variable-width font

> It would also be helpful to have a font size setting for the content areas, and maybe bump up the default font size for the content just a bit.

I came looking to open an issue on exactly these items. It's much harder to read after the update for me and I'd love more control over the font used. At the very least a "small, medium, large" setting for size but a monospace font is also greatly desired.

markerikson:
Related: the "Stack Trace" pane seems to have lost any formatting as well:

![image](https://user-images.githubusercontent.com/1128784/149675465-96d5d3f0-f8d6-4d4d-b818-118b9c88f4b8.png)

Methuselah96:
That looks the same (except for the font which I expect to fix soon) as 2.17:
![image](https://user-images.githubusercontent.com/693755/149680810-8753e244-8150-4c9f-8555-964d3c148c4e.png)

Methuselah96:
[...]

PR #996, opened 2022-01-08
fix(rtk-query-monitor): capitalize tabs #995

FaberVitale:
Partially fixes https://github.com/reduxjs/redux-devtools/issues/995

---

@Methuselah96  Is there a way make VScode work with plug'n'Play:
It is a squiggly line fest and not even the good ol' `restart TS server` seems to work in this case.


Methuselah96:
> @Methuselah96 Is there a way make VScode work with plug'n'Play: It is a squiggly line fest and not even the good ol' `restart TS server` seems to work in this case.

Looks like it might [require a plugin](https://yarnpkg.com/features/pnp#support-via-plugins). Sorry about that.

PR #1003, opened 2022-01-11
Fix Symbol.observable polyfill mismatch

Methuselah96:
Fixes https://github.com/reduxjs/redux-devtools/issues/1002. The Symbol.observable was ponyfilling to different values depending on whether `symbol-observable` had been imported yet. This fixes it by waiting to call the ponyfill until the store is being created. This should make sure that it is consistent with the symbol used by Redux.


Methuselah96:
Forgot about `subscribe`. Will update shortly.

PR #1031, opened 2022-01-18
Remove React from page bundle

Methuselah96:
Fixes https://github.com/reduxjs/redux-devtools/issues/1030. The issue is that we're importing `persistState` from `@redux-devtools/core` in the page bundle, but that ends up importing all of `@redux-devtools/core` which imports `react`.

This draft PR is just to get unstuck until I figure out how this should be done without copy-pasting `persist-state` from `@redux-devtools/core`. Looks like multiple entry points using the exports field in `package.json` is still being worked on for TypeScript (or at least the issue for it is still open) and I'm not really sure if the methods in `@redux-devtools/core` can be considered "side-effect free." They definitely have side-effects, but the side-effects aren't meaningful to anybody else and should be fine to be tree-shaken.

tl;dr need to investigate more but can publish this for now

### DIFF
--- a/extension/src/app/stores/enhancerStore.ts
+++ b/extension/src/app/stores/enhancerStore.ts
@@ -1,6 +1,6 @@
 import { Action, compose, Reducer, StoreEnhancerStoreCreator } from 'redux';
 import { instrument } from '@redux-devtools/instrument';
-import { persistState } from '@redux-devtools/core';
+import persistState from './persistState';
 import { ConfigWithExpandedMaxAge } from '../../browser/extension/inject/pageScript';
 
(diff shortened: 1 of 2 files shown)


timdorr (approved):

timdorr:
This looks good to me. Do we want to push this out ASAP?

Methuselah96:
Yes, it has been published, just waiting for the Chrome Web Store to approve it.

timdorr:
Well, let's get this merged so the code matches what's going out in the extension.

martinmckenna:
very confusing that the extension is being published from open PRs

PR #1049, opened 2022-01-23
Remove UNSAFE_componentWillReceiveProps from react-json-tree

defunctzombie:
Reworking of https://github.com/reduxjs/redux-devtools/pull/644 to remove UNSAFE_componentWillReceiveProps which is deprecated in react 17 and likely removed in 18.

Components that used UNSAFE_componentWillReceiveProps are converted to functional components with hooks.


packages/react-json-tree/src/JSONNestedNode.tsx
```diff
@@ -97,144 +97,116 @@ interface Props extends CircularPropsPassedThroughJSONNestedNode {
...
-    if (getStateFromProps(this.props).expanded !== nextState.expanded) {
-      this.setState(nextState);
+  // When certain props change, we need to re-compute whether our node should be in an expanded state
+  useEffect(() => {
+    setExpanded(() => {
+      return !isCircular ? shouldExpandNode(keyPath, data, level) : false;
+    });
+  }, [isCircular, data, keyPath, level, shouldExpandNode]);
+
+  // fixme - previously this was happening after a component should update
```
defunctzombie, line 130: @Methuselah96 Are you familiar with this renderChildNodes code? I believe the intent is that it doesn't run on ever render but only when something has changed. I'm guessing not when all props have changed but only some specific props?

Methuselah96: It's a little bit complicated because the components can be rendered recursively. For example:

JSONTree -> JSONNode -> JSONObjectNode -> JSONNestedNode -> renderChildNodes -> JSONNode -> ...

That means that `renderChildNodes` does need to run for the majority of prop changes because most of the props get passed down the hierarchy recursively. You could try to create a list of only the props that are recursively passed through the children based on the [circular props type](https://github.com/reduxjs/redux-devtools/blob/c379a6208119aaef5629bddfb3cb1aa447c6a1f4/packages/react-json-tree/src/types.ts#L68-L71), however I would lean toward just calling `renderChildNodes` on every render.

packages/react-json-tree/src/index.tsx
```diff
@@ -47,133 +46,56 @@ const defaultLabelRenderer = ([label]: (string | number)[]) => (
 );
 const noCustomNode = () => false;
 
-function checkLegacyTheming(theme: Theme | undefined, props: Props) {
```
defunctzombie, line 50: @Methuselah96 As part of this update I removed support for the deprecated styling methods. Removing these is not strictly required to do in this change (just something that made the code easier to reason about). Since this could be seen as a breaking change I am happy to put this back if desired.

packages/react-json-tree/src/JSONNestedNode.tsx
defunctzombie: 👍 Removed the fixme comment - this code now runs on every render

Methuselah96:
@defunctzombie  Looks like there are some type errors:
```
Error: src/JSONArrayNode.tsx(21,4): error TS2322: Type '{ data: any; nodeType: string; nodeTypeIndicator: string; createItemString: (data: any) => string; expandable: boolean; styling: StylingFunction; keyPath: (string | number)[]; ... 12 more ...; children?: ReactNode; }' is not assignable to type 'Pick<Props, "nodeType" | "styling" | "keyPath" | "labelRenderer" | "valueRenderer" | "circularCache" | "level" | "isCircular" | "shouldExpandNode" | "hideRoot" | "getItemString" | ... 6 more ... | "expandable">'.
  Types of property 'circularCache' are incompatible.
    Type 'any[] | undefined' is not assignable to type 'any[]'.
      Type 'undefined' is not assignable to type 'any[]'.
Error: src/JSONIterableNode.tsx(33,6): error TS2741: Property 'expandable' is missing in type '{ nodeType: string; nodeTypeIndicator: string; createItemString: (data: any, limit: number) => string; data: any; styling: StylingFunction; keyPath: (string | number)[]; ... 12 more ...; children?: ReactNode; }' but required in type 'Pick<Props, "nodeType" | "styling" | "keyPath" | "labelRenderer" | "valueRenderer" | "circularCache" | "level" | "isCircular" | "shouldExpandNode" | "hideRoot" | "getItemString" | ... 6 more ... | "expandable">'.
Error: src/JSONObjectNode.tsx(20,4): error TS2322: Type '{ data: any; nodeType: string; nodeTypeIndicator: string; createItemString: (data: any) => string; expandable: boolean; styling: StylingFunction; keyPath: (string | number)[]; ... 12 more ...; children?: ReactNode; }' is not assignable to type 'Pick<Props, "nodeType" | "styling" | "keyPath" | "labelRenderer" | "valueRenderer" | "circularCache" | "level" | "isCircular" | "shouldExpandNode" | "hideRoot" | "getItemString" | ... 6 more ... | "expandable">'.
  Types of property 'circularCache' are incompatible.
    Type 'any[] | undefined' is not assignable to type 'any[]'.
      Type 'undefined' is not assignable to type 'any[]'.
```

defunctzombie:
@Methuselah96 I looked at the types here and can't make sense of where this error comes from or what the intended behavior is. I did not change any of the Prop types in this PR.

The error comes from a miss-match between these two types:

https://github.com/reduxjs/redux-devtools/blob/main/packages/react-json-tree/src/types.ts#L54
https://github.com/reduxjs/redux-devtools/blob/main/packages/react-json-tree/src/types.ts#L65

One being optional for `circularCache`.

Methuselah96:
Previously, `circularCache` was [defaulted to an empty array](https://github.com/reduxjs/redux-devtools/blob/main/packages/react-json-tree/src/JSONNestedNode.tsx#L137) in `JSONNestedNode` if it wasn't provided, however it looks like this PR does not initialize the circular cache (and perhaps also failing to update it?).

defunctzombie:
Even if I updated JSONNestedNode to set a default for circularCache, that doesn't prevent the error. The issue is that JSONArrayNode props are of type `CircularPropsPassedThroughJSONNode` which extend `JSONNestedNodeCircularPropsPassedThroughJSONNode` which has `circularCache?: any[];` however JSONNestedNode props are `CircularPropsPassedThroughJSONNestedNode` which extend `JSONNestedNodeCircularPropsPassedThroughJSONNestedNode` which has `circularCache: any[];`.

So when JSONArrayNode does `<JSONNestedNode {...props} ...etc` circularCache types don't match since JSONNestedNode doesn't allow undefined for circularCache. Its possible that the defaultProps from before is treated as turning `circularCache: any[]` into `circularCache?: any[]` even tho the Props type does not allow circularCache to be undefined.

The two fixes are:

- add `circularCache={props.circularCache ?? emptyArray}` to JSONArrayNode's use of JSONNestedNode
- update the Props for JSONNestedNode to allow circularCache to be optional

Preferences?

Methuselah96:
Right, if a prop is specified in `defaultProps` then TypeScript will make that prop optional when the component is used.

Option 1 is not preferable because `JSONArrayNode` is not the only component that uses `JSONNestedNode` (`JSONIterableNode` and `JSONObjectNode` do as well). Option 2 sounds good to me (accompanied with actual defaulting `circularCache` to an empty array and updating it appropriately).

defunctzombie:
I've updated the typescript prop types to reflect allowing undefined.

packages/react-json-tree/src/JSONNestedNode.tsx
```diff
@@ -94,147 +94,116 @@ interface Props extends CircularPropsPassedThroughJSONNestedNode {
...
-  UNSAFE_componentWillReceiveProps(nextProps: Props) {
-    const nextState = getStateFromProps(nextProps);
-    if (getStateFromProps(this.props).expanded !== nextState.expanded) {
-      this.setState(nextState);
+  // When certain props change, we need to re-compute whether our node should be in an expanded state
+  useEffect(() => {
+    setExpanded(() => {
+      return !isCircular ? shouldExpandNode(keyPath, data, level) : false;
+    });
+  }, [isCircular, data, keyPath, level, shouldExpandNode]);
```
Methuselah96, line 127: I believe this would reintroduce the undesired behavior discussed [here](https://github.com/alexkuz/react-json-tree/pull/61#issuecomment-280425149) since `shouldExpandNode` is often intended to only indicate whether on a node should be expanded on the first render. This was fixed by evaluating `shouldExpandNode` for the previous props and the current props and only updating the expanded state if they differ.

We're unfortunately stuck in an awkward situation where this component is partially controlled and it's unclear when `shouldExpandNode` should be called. We either need to preserve the previous behavior (which seems difficult if we don't want to keep around a reference to the previous props) or come up with a better solution that avoids the difficulties that arise from a prop that allows for partially controlled state. Feel free to pitch some ideas so we can discuss before putting too much working into a single solution. I'll keep thinking about it myself.

Methuselah96 (changes requested):
Thanks for taking this on! I think there are a few things that still need to be addressed before this can be merged. Also, can you add a changeset that includes a description of the change and notes the breaking change of removing the legacy theming props?

unematiii:
@Methuselah96 @defunctzombie Is any of you still working on it?

Methuselah96:
I am not currently working on it.

defunctzombie:
I too am not currently working on this.

Methuselah96:
It has not been resolved yet, but I am working on [a fix](https://github.com/reduxjs/redux-devtools/pull/1288) that I hope will be released within the next week.

PR #1051, opened 2022-01-24
Default monitors to a monospace font

Methuselah96:
Fixes https://github.com/reduxjs/redux-devtools/issues/995.

### DIFF
--- a/packages/redux-devtools-app/src/containers/monitors/Slider.tsx
+++ b/packages/redux-devtools-app/src/containers/monitors/Slider.tsx
@@ -27,7 +27,7 @@ class Slider extends Component<Props> {
   }
   render() {
     return (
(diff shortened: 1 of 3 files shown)


Methuselah96:
2.x Inspector Monitor:
inspector-old

Inspector Monitor as of this PR:
image

RTK Query Monitor as of this PR:
image

This PR goes a little bit overboard with making too many things monospace. It seems that ideally there would be some more refactoring of how styling works in the existing monitors to use `@redux-devtools/ui` which would provide a uniform way to declare which font type to use in different situations. However currently only the slider monitor uses `@redux-devtools/ui`. Given that this seems like an improvement over the status quo, I'm planning to merge and release this and the work of standardizing our approach can be tackled in a future refactor.

PR #1094, opened 2022-02-21
Update branch name on package file

keponk:
* License file is extremely useful for crawlers looking for 3rd party software libs. I've used the exact same one from other repos under reduxjs.
* `package.json` file was still pointing at `master`, instead of `main` which is the current main branch.


Methuselah96 (approved):

Methuselah96:
Thanks!

PR #1130, opened 2022-04-08
Improved RTK Query monitor to focus more on data.

### LINKED ISSUE #1126: Improve RTK Query monitor to focus more on data
Quoting user feedback in https://github.com/reduxjs/redux-toolkit/discussions/2206 :

> I've gotten used to using the "state" tab in the redux dev tools, where I can see the current state at a glance. However, with the rtk query dev tools, I need to click through multiple nested objects in order to get to the data field for a query, meaning that I often resort to console.logging the data rather than using the dev tools. I realize that technically rtk query saves lots of other types of state, but most of the time, I'm only interested in the `data` that is actually being used in my app.

Here's the current UI:

image

I think it might be worth having, say, "Data" and "Cache Entry" (TBD?) tabs.  One way or another, the actual data _should_ be immediately visible as soon as you click on the request entry in the list.

Tagging @FaberVitale for further thoughts.

Bezmehrabi:
closes #1126.

the UI is:
rtk-data
rtk-data-full


Methuselah96:
Sorry for the confusion on this, has been resolved by https://github.com/reduxjs/redux-devtools/pull/1129.

PR #1194, opened 2022-06-23
Fix ReduxDevtools to work in Sizzy

Zeko369:
[Sizzy](https://sizzy.co) is an electron powered browser that mocks UserAgent to match the simulated device, so ReduxDevtools can't detect it as Electron.

Sizzy does however expose a global `__sizzy` attribute with which we can detect it's running in Sizzy


Methuselah96:
Is this still an issue? It would be nice if Redux DevTools didn't have to know about Sizzy, and it seemed like maybe [that was being worked on by Sizzy](https://twitter.com/thekitze/status/1562861831999614977).

francisprovost:
It is still an issue yes.

If this PR is the best solution I couldn't say but IMO I see this in the same way as the `isFirefox` but it's not my call to make.

Zeko369:
Fixed upstream in Sizzy

PR #1201, opened 2022-07-19
Provide explicit constraints for TypeScript 4.8

DanielRosenwasser:
Hi there! 👋

TypeScript recently added some stricter checking around unconstrained generics being incompatible with emptyish object types. The changes are only in nightly versions of TypeScript right now, but we've been exploring how open-source projects have been impacted. These are the changes necessary for redux-devtools-extension to build cleanly.


Methuselah96 (approved):
Thanks!

PR #1235, opened 2022-10-05
Fix link to paragraph 1.3

Tiketti:
A straight-forward change: the link to paragraph 1.3 ("Use @redux-devtools/extension package from npm") within the same document was not working. This fixes it.


Methuselah96 (approved):
Thanks!

PR #1237, opened 2022-10-07
Update Troubleshooting.md

eghernqvist:
Added documentation for another possible cause of "### It shows only the `@@INIT` action or moving back and forth doesn't update the state".


Methuselah96 (approved):
Thanks!

PR #1305, opened 2022-12-28
feature(redux-devtools-app/ui): add search feature to state

### LINKED ISSUE #989: Search for entities
We are using the devtools in our project and they are really helpfull. But one thing is missing from my point of view: The possibility  to search for an entity. We have e.g. one EntityState with roundabout 600 entities. It would be very helpfull if I could just enter the ID of a specific one to find it faster.

80avin:
Add search feature for State tree.

Search is done using a fork of [JSONPath-plus](https://www.npmjs.com/package/jsonpath-plus) (fork used: https://github.com/JSONPath-Plus/JSONPath/pull/185 ) which adds support for Content Security Policy by eliminating the use of `eval/Function`.

Being JSONPath, we can use complex expressions to search the tree by property names or values.

A sample expression is `$..[?(@ && @.id%3 === 0)]` tried in https://zalmoxisus.github.io/examples/router/#/Standard%20Todo?_k=pnapms

Fixes https://github.com/reduxjs/redux-devtools/issues/989


80avin:
@ximfritz I couldn't get time to complete the PR in JSONPath-plus.
But partly, I've also been lazy since I had made this feature usable on my local, only cleanup, writing tests & other PR formalities were left.

I'll try to finish it in few weeks now.

PR #1365, opened 2023-03-06
Fix responsive layout

Methuselah96:
Resolves https://github.com/reduxjs/redux-devtools/issues/1294.


Methuselah96:
@FaberVitale FYI, this reverts the changes made in https://github.com/reduxjs/redux-devtools/pull/1129 to bump the min width if you want to take another stab at that.

PR #1467, opened 2023-08-23
fix: view on popup

agungfir10:
This PR is change style on popup, in my problem is view on popup very disgust...

Problem

Solved:
Popup Redux is Right

### DIFF
--- a/extension/src/style.pug
+++ b/extension/src/style.pug
@@ -7,7 +7,7 @@ style.
     overflow: hidden;
     height: 100%;
     width: 100%;
-    min-width: 350px;
+    min-width: 720px;
     min-height: 400px;
     margin: 0;
     padding: 0;


Methuselah96 (changes requested):
I believe doing this would break the responsive layout as demonstrated in https://github.com/reduxjs/redux-devtools/issues/1294, which was why the `min-width` was reverted back to `350px` in https://github.com/reduxjs/redux-devtools/pull/1365. Can you verify this doesn't break the responsive layout?

extension/src/style.pug
```diff
@@ -7,7 +7,7 @@ style.
     overflow: hidden;
     height: 100%;
     width: 100%;
-    min-width: 350px;
+    min-width: 720px;
```
batya-bicha, line 10: maybe a 760px would be better, like it used to be?

Methuselah96: To be clear, the popup used to size to content which was almost always bigger than `min-width`. I suspect a change in Chrome 116 makes it so it no longer resizes to fit the content.

Methuselah96:
Superseded by https://github.com/reduxjs/redux-devtools/pull/1468.

PR #1487, opened 2023-09-13
Replace deprecated action

nickmccurdy:
GitHub's default Node version doesn't require additional downloads or installations. Currently the LTS and default version are similar, but I think it would be best to stay consistent with the other repositories.


Methuselah96:
Thanks, I see the PRs you made in the other Redux repositories. I think redux-devtools is separate enough that it doesn't need to match the other repositories and using the LTS version doesn't pose any security concerns (which seems to be the motivating factor for the other PRs).

As a preference, I think I'd rather just stick with LTS.

nickmccurdy:
Just because the LTS and default version are similar now, doesn't mean they'll be the same in the future. @acemarke has told me he doesn't mind which Node versions are used, and using the default version also means it's always installed and should result in faster builds.

Methuselah96:
Node LTS is a much more firm/established version than the default GitHub Actions Node version which is why I'd rather stick with that. A significant divergence between the two seems quite hypothetical, so I'd rather just deal with that if/when it comes up.

The build time is at 14 minutes, I don't think using the default Node version will meaningfully speed up the build (happy to hear evidence to the contrary).

Surely this is not a big deal? I see somewhat of a reason to update the other Redux repositories to only use supported versions of Node, but it seems like sticking with LTS should be fine here?

nickmccurdy:
I'd rather stay consistent with what I discussed with Mark, unless he changes his mind.

Also it wasn't necessarily my intention for this PR to only change default versions. I've renamed it to reflect that I'm also fixing deprecated workflows.

Methuselah96:
Thanks, the deprecation fixes are definitely appreciated and worth keeping.

The Redux maintainers have given me a certain level of autonomy with redux-devtools. I believe this is something they're fine with me making a judgment call on. Since redux-devtools is maintained fairly autonomously from the other Redux packages, it's not something that needs to be 100% consistent with the other Redux repos. My preference is to stick with LTS (for the reasons given above) which should be fine from a security perspective.

markerikson:
Let's go with Nathan's preferences here.

Methuselah96 (approved):
Thanks!

PR #1622, opened 2024-03-04
Adds argument to prevent Redux DevTool CLI disconnecting when debugging

jardicc:
Adds new CLI argument to change pingTimeout value. This solves few issues for me if I set crazy high number.

1) UXP in Photoshop has UWP implementation of WebSockets and its client when asked to send empty message it is never sent but everything else is. Thefore it gets disconnected. https://github.com/reduxjs/redux-devtools/issues/1614

2) I think for some older middleware using protocol V1 sending different ping content this could help as well https://github.com/zalmoxisus/remote-redux-devtools/issues/158

3) When you debug client app and you stop on breakpoint for 20s which default value of socket cluster... it will disconnect debugger as well because you stopped client app so it can't send a ping. So value can be increased to prevent that.

Also `maxRequestBody` was wrong.

Please let me know if I should change something or feel free to change it yourself. I was thinking about seconds instead of `ms` but not sure what is better.

### DIFF
--- a/packages/redux-devtools-cli/CHANGELOG.md
+++ b/packages/redux-devtools-cli/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+## 4.1.0
+
+### Minor changes
+
+- adds CLI argument to change ping timeout `--pingTimeout=20000` to make debugging easier
+- fixes `--maxRequestBody` argument
+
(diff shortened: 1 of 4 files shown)


jardicc:
Can someone please review this PR?

PR #1635, opened 2024-03-15
Update out-of-date documentation link

ChrisdeWolf:
Old link was resolving to a 404 (not found). Updated to what i think was the updated url/most helpful given the context.


Methuselah96 (approved):
Thanks!

PR #1655, opened 2024-04-15
Move the logic from @redux-devtools/app into @redux-devtools/app-core

matt-oakes:
This PR follows on from a previous discussion: https://github.com/reduxjs/redux-devtools/discussions/1653

This change splits out the main logic from the Redux Devtools App into a new core package but keeps the socket connection management in @redux-devtools/app. The aim is to allow for easier reuse of the rest of the app in other envioronments with their own transport methods, such as React Native or Electron.

The goal of this PR is that nothing will change for users of the `@redux-devtools/app` package. It re-exports all of the exports from `@redux-devtools/app-core` so everything that was previously available still is.

The majority of the PR is just moving files and changes to imports. Some changes needed to be made to allow for some actions, reducers, and middleware to be kept in `@redux-devtools/app` rather than moved to `@redux-devtools/app-core`. These are anything which relates to the management of the socket connection.

There is also a new prop on the `App` component which allows adding additional settings panels. This is used to add the "Connection" settings component, but not include it in the core package as it's specifically socket related. This poperty is optional to avoid any breaking changes.


matt-oakes:
I have made the changes needed for the linting and tests to pass correctly. I've also added a changeset file which will set the new `app-core` package to `1.0.0` and do a minor upgrade on the `app` package as it's not a breaking change.

You should have push access, so feel free to make changes or comments as required.

matt-oakes:
Whoops! I fixed the small formatting issue. Sorry, my editor had a different way to automatically format that and I forgot to run the check formatting locally.

matt-oakes:
To test this out I have created [an Expo Dev Plugin](https://github.com/matt-oakes/redux-devtools-expo-dev-plugin/tree/main) which makes use of this new `app-core` package (temporarily published to NPM as `@mattoakes/redux-devtools-app-core`).

It's working great and the amount of code needed to get the full devtools app UI with a custom connection middleware is now greatly reduced.

Methuselah96 (approved):
Thanks, let's give this a try!

matt-oakes:
Thank you! Please let me know if there are any issues or anything you need to discuss.

Methuselah96:
@matt-oakes The `@redux-devtools/app-core` package should be live now, let me know if you run into any issues.

arstropica:
> Thank you! Please let me know if there are any issues or anything you need to discuss.

[Themes.tsx](https://github.com/reduxjs/redux-devtools/blob/main/packages/redux-devtools-app-core/src/components/Settings/Themes.tsx) could previously be imported at `@redux-devtools/app/lib/esm/components/Settings/Themes` but access now appears to be limited internally. Is there another method of accessing the component?

PR #1658, opened 2024-05-06
Enable lock button for non-Redux connections.

EskiMojo14:
Currently the lock button is disabled if the app doesn't think it's talking to a Redux store - which is decided here: https://github.com/reduxjs/redux-devtools/blob/f019024b21e119f6e123bd1f9e2d815d295dfde6/packages/redux-devtools-app/src/reducers/instances.ts#L280

There doesn't seem to be a reason stated for this, and stores other than Redux can certainly implement locking, so I think it would be good to enable this. It's still hidden based on the feature option, which defaults to only being true if it's a Redux store: 
https://github.com/reduxjs/redux-devtools/blob/f019024b21e119f6e123bd1f9e2d815d295dfde6/packages/redux-devtools-app/src/reducers/instances.ts#L290

### DIFF
--- a/.changeset/neat-ducks-do.md
+++ b/.changeset/neat-ducks-do.md
@@ -1,5 +1,5 @@
 ---
-"@redux-devtools/app": minor
+'@redux-devtools/app': minor
 ---
 
 Enable lock button for non-Redux connections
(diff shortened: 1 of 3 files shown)


faiyun123 (commented):
****

PR #1692, opened 2024-06-21
Avoid auto-selecting instance in iFrames

### LINKED ISSUE #1000: Instance auto-switches undesirably
I have an app that has an iframe within. There's a redux instance on the top frame and another instance (different app) inside the iframe. The tools have the instance switcher dropdown which shows 3 values: "autoselect instance", "instance 1" and "instance 2". 

No matter what's selected on the instance switcher, when ever an action happens on the top frame instance (i.e. instance 1), the tools go back to showing/selecting that instance — even when the action occurs on the instance 2. 

It doesn't switch to instance 2 from instance one ever, only from 2 to 1 for all events.

It gets very disruptive when you're trying to debug on instance 2. It feels a lot like a bug.

alexandcote:
**Note: First time contributing to this project, open to feedback**

Avoid auto-selecting the outer frame instance when redux exists in iFrame embedded on the page.

Fix: #1000

iFrame has a different tabID than the main frame. The connections are therefore not grouped properly.

Example: 
```
{
  "connections": {
    "<tabId>" ["<firstConnectionID>"],
    "<iFrameTabId>": ["<secondConnectionID>"]
  }
}
```

Should we group those connections under the same key? 
The fix is working but I'm unsure if it's the right solution to the problem.

### DIFF
--- a/.changeset/fluffy-steaks-promise.md
+++ b/.changeset/fluffy-steaks-promise.md
@@ -1,0 +1,5 @@
+---
+'remotedev-redux-devtools-extension': patch
+---
+
+Avoid auto-selecting instance with iFrames
--- a/extension/src/window/store/instanceSelectorMiddleware.ts
+++ b/extension/src/window/store/instanceSelectorMiddleware.ts
@@ -13,7 +13,16 @@ function selectInstance(
 ) {
   const instances = store.getState().instances;
   if (instances.current === 'default') return;
-  const connections = instances.connections[tabId];
+
+  const connections = Object.entries(instances.connections).reduce<
+    (string | number)[]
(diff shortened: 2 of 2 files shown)


Methuselah96:
Hello @alexandcote, sorry for the lack of response so far. I've been digging into the code that controls auto-selection since some of it is related to [upgrading to Manifest V3](https://github.com/reduxjs/redux-devtools/issues/1037). As part of that investigation, I believe I've found a fix for this bug and I expect it will be included in the next release, which I hope to put out in the next few weeks.

Sorry it has taken so long to get to this, but I hope this will be fixed soon.

markerikson:
Should be resolved by #2117

PR #1752, opened 2024-09-05
Fix action missing bug when playback speed is set to live

### LINKED ISSUE #1751: Replaying on speed `live` causes the devtools to go blank
Tested on Mac on Chrome and Firefox.

Steps to repro:
Switch the speed to live from 1x.
Click play.
Once the playback is complete, the entire Redux DevTools will go blank. You have to open/close the tool to get it to work again.

xerebz:
This fixes #1751. I've tested it by building it locally and running the extension on Chrome.


xerebz:
Force pushed change. Needed to run prettier.

PR #1934, opened 2025-09-21
Fix CSP 'unsafe-eval' error preventing Remote DevTools connection

sangmokh:
# Issue

https://github.com/reduxjs/redux-devtools/issues/1783

image

# Analysis

The `unsafe-eval` CSP error occurs in [`Form.tsx`](https://github.com/reduxjs/redux-devtools/blob/main/packages/redux-devtools-ui/src/Form/Form.tsx) when `@rjsf/validator-ajv8` attempts to use `eval()` for JSON schema validation. This happens because the library falls back to runtime compilation when [precompiled validators](https://rjsf-team.github.io/react-jsonschema-form/docs/usage/validation/#precompiled-validators) are not available.

# Solution / Workaround

Addressing the root cause by migrating to precompiled validators would require significant refactoring, as `Form.tsx` is used with multiple schema configurations across multiple components. Yet the Remote Redux DevTool feature is unsuable without fixing the form error.

Two simple workarounds were considered:
- Disable form valuation of the Connection Setting form
- Update the CSP in [remote.pug](https://github.com/reduxjs/redux-devtools/blob/main/extension/src/remote/remote.pug) to allow `unsafe-eval`

This PR opted for the first approach because the UI already constrains the `port` field to numeric values and the `secure` field to boolean values, making the `eval`-based ajv validation redundant. The second approach was avoided as it would compromise the security of the Remote DevTools window.


msanchezdev:
Waiting for this as well

